PR #1, opened 2017-05-31
describe zync

mikz:


README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Before describing the design it is important to define the terms.
+
+* **Zync** - deployment of this project, Web API.
+
+
+* **3scale** - 3scale API Manager (system).
+* **Tenant** - mapping of Provider id registered in 3scale API Manager to domain and access token.
+* **Model** - relevant object in **3scale** like Application, Limit, Metric, Service.
+* **Notification** - Message sent to **Zync** descibing the **model** that changed and having all required properties to fetch it from the API later.
```
davidor: typo: describing

README.md
```diff
@@ -0,0 +1,41 @@
+# Zync - the sync tool
+
+Zync is going to take your 3scale data and push them somewhere else, reliably.
```
kevprice83: push **it** somewhere else

README.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+* **3scale** - 3scale API Manager (system).
+* **Tenant** - mapping of Provider id registered in 3scale API Manager to domain and access token.
+* **Model** - relevant object in **3scale** like Application, Limit, Metric, Service.
+* **Notification** - Message sent to **Zync** descibing the **model** that changed and having all required properties to fetch it from the API later.
+  * Example: Application 3456, tenant_id: 26
+  * Example: Limit 4566, metric_id: 36, application_id: 46, tenant_id: 16
+* **Update** - **Zync** fetches updated **Model** from the **Notifier** via the API using the information provided by the **Notification**.
+* **Lock** - mechanism that prevents two pieces of code to be executing in parallel for the same scope
```
unleashed: I would rephrase as protecting/preventing concurrent data access, which is the goal you want them for (as oppose to executing parallel -concurrent?- code).

README.md
```diff
@@ -0,0 +1,41 @@
...
+  * Example: **Tenant Lock** would mean only one can be running for one **Tenant**
+  * Example: **Model** **Lock** - only one per uniquely identified **Model**
+* **Entry** - The information from the API provided by the **Update**.
+* **Log** - ordered list of **Entries** as they were received
+* **Integration** - code that pushes **Log** **entries** one-by-one for the same **Model** to some external service
+
+## Design
+
+**Zync** is meant to synchronize data from **3scale** to external systems (like IDPs). Some people use for this Webhooks but without further logic they can be unreliable and arrive out of order. This tool is meant to synchronize the final state into different system.
```
kevprice83: Some people use Webhooks for this....into different system(s)

unleashed: grammatical suggestions :) - `Some people use Webhooks for this` ... `state to a different system`

mikz: Geez, wtf I wrote 🙂 Thx

README.md
```diff
@@ -0,0 +1,41 @@
...
+* **Notification** - Message sent to **Zync** descibing the **model** that changed and having all required properties to fetch it from the API later.
+  * Example: Application 3456, tenant_id: 26
+  * Example: Limit 4566, metric_id: 36, application_id: 46, tenant_id: 16
+* **Update** - **Zync** fetches updated **Model** from the **Notifier** via the API using the information provided by the **Notification**.
+* **Lock** - mechanism that prevents two pieces of code to be executing in parallel for the same scope
+  * Example: **Tenant Lock** would mean only one can be running for one **Tenant**
+  * Example: **Model** **Lock** - only one per uniquely identified **Model**
+* **Entry** - The information from the API provided by the **Update**.
+* **Log** - ordered list of **Entries** as they were received
+* **Integration** - code that pushes **Log** **entries** one-by-one for the same **Model** to some external service
```
unleashed: There is no word about model relationships, or more accurately, model dependencies. I think we should make it explicit (or at least leave a note open to implementation design when the need/problem arises).

README.md
```diff
@@ -0,0 +1,41 @@
...
+
+The flow is defined as **3scale** -> **Zync** ( <- **3scale**) -> **Integration**. So **3scale** notifies **Zync** there was a change to a **model** but does not say more than primary key and information required to fetch it from the **3scale** API.
+
+**Zync** upon receiving the notification will acquire an **update model lock** and try to perform an **update**. Any information received this way is added as an **entry** to the **log** and the **model lock** is released. That **entry** can be either new data or information that the record is no longer there (404 from the API). If new **notification** came when the **model lock** was acquired, it is going to be processeed after the lock is released.
+
+After adding **entry** to the **log** an **integration** is triggered and aquires an **integration model lock** so it will process only one (latest) **entry** for a **model** at the time. After the **integration** finishes (with both failure or success) it will release the lock and trigger another run if failed.
```
kevprice83: at a time

README.md
mikz: Yep, thanks!

README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Properties
+
+Given the locking on the **model** there will be some parallelisation, but also updates to one object will be serialized. This needs to be done to limit the network issues and ensure the request was delivered before issuing new one. 
+
+Because **Zync** will keep a **log** of **events** it will be able to replay changes and recover last state just taking last revisions of each **model** and even remove the ones that been created before but have been deleted.
```
kevprice83: that were created

README.md
mikz: Yep. Like: integration can access the **entries** **log** to fetch more data and for example handle **model** dependencies by accessing all dependent **models** and delete them before deleting the parent.

mikz:
@kevprice83 this is one directional sync (probably good to mention it explicitly). So no changes from external systems are going to be propagated to 3scale. 

In case of IDPs that should not matter because 3scale does not need any of the information. Gateway does and will be connecting to the IDP directly and pulling the information when needed.

Thanks for the corrections!

kevprice83:
@mikz okay that makes sense. I assume the gateway will be removed from the whole OAuth handshake in the IdP integration then (Sorry I know that is another topic)...

mikz:
@kevprice83 exactly, this does the synchronisation, so the Gateway can use the IDP directly.

README.md
unleashed: forgot about this one @mikz

mikz: damn, I used spellchecker! but probably did not start from the beginning 🙂

mikz: fixed in 03a91a0

README.md
```diff
@@ -0,0 +1,41 @@
...
+
+The flow is defined as **3scale** -> **Zync** ( <- **3scale**) -> **Integration**. So **3scale** notifies **Zync** there was a change to a **model** but does not say more than primary key and information required to fetch it from the **3scale** API.
+
+**Zync** upon receiving the notification will acquire an **update model lock** and try to perform an **update**. Any information received this way is added as an **entry** to the **log** and the **model lock** is released. That **entry** can be either new data or information that the record is no longer there (404 from the API). If new **notification** came when the **model lock** was acquired, it is going to be processed after the lock is released.
+
+After adding **entry** to the **log** an **integration** is triggered and acquires an **integration model lock** so it will process only one (latest) **entry** for a **model** at a time. After the **integration** finishes (with both failure or success) it will release the lock and trigger another run if failed.
```
macejmic: how many times it will re-run failed entry?

README.md
```diff
@@ -0,0 +1,41 @@
...
+* **Entry** - The information from the API provided by the **Update**.
+* **Log** - ordered list of **Entries** as they were received.
+* **Integration** - code that pushes **Log** **entries** one-by-one for the same **Model** to some external service.
+  * Integration can access the **entries** **log** to fetch more data and for example handle **model** dependencies by accessing all dependent **models** and delete them before deleting the parent.
+
+## Design
+
+**Zync** is meant to synchronize data from **3scale** to external systems (like IDPs). Some people use Web-hooks  for this but without further logic they can be unreliable and arrive out of order. This tool is meant to synchronize the final state to a different systems.
+
+The flow is defined as **3scale** -> **Zync** ( <- **3scale**) -> **Integration**. So **3scale** notifies **Zync** there was a change to a **model** but does not say more than primary key and information required to fetch it from the **3scale** API.
```
macejmic: so it's gonna be primary key + api endpoint?

README.md
mikz: I think that is very good point and we should keep status.

I'll describe in an update that there is also something like **status** for every **model** that keeps information about last incoming **notification** and last outgoing **push** with number of retries and few other metadata.

README.md
mikz: For some models it is primary key for some it is few more metadata. You'll need service_id and metric_id to limit from the API.

PR #29, opened 2017-06-23
integrate license finder

mikz:
to ensure we use only free open source software

### DIFF
--- a/doc/dependency_decisions.yml
+++ b/doc/dependency_decisions.yml
@@ -41,3 +41,10 @@
     :why: 
     :versions: []
     :when: 2017-06-23 06:16:00.472944000 Z
+- - :license
+  - bundler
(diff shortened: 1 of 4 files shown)


doc/dependency_decisions.yml
```diff
@@ -0,0 +1,43 @@
...
+    :when: 2017-06-23 06:12:33.037106000 Z
+- - :whitelist
+  - Apache 2.0
+  - :who: 
+    :why: 
+    :versions: []
+    :when: 2017-06-23 06:12:54.025599000 Z
+- - :whitelist
+  - Simplified BSD
+  - :who:
```
mikz: Let me know if you need these comments or we can generally accept that all BSD/MIT licenses are approved.

PR #41, opened 2017-08-08
integrate message_bus

mikz:
so other components can consume status of zync


mikz:
OpenShift HAProxy Router works correctly with MessageBus chunked long polling.

PR #44, opened 2017-08-16
xml report for license finder

### LINKED ISSUE #43: Generate license.xml THREESCALE-170
Part of https://issues.jboss.org/browse/THREESCALE-170
We need to generate a licenses.xml for packages installed.

Basically it should be a rake task that will use https://github.com/3scale/license_finder_xml_reporter

Martouta:
will genertate xml licenses for Zinc after running a rake command `$ bundle exec license_finder`

closes #43


jamaga:
to generate licences 
`bundle exec license_finder`

jamaga:
@ddcesare   we are missing some licenses and donno if maybe u will catch what is maybe missing

when we do `grep packageName doc/licenses/licenses.xml| wc -l` we get 53
and when `bundle list | wc -l `   - 87

lib/tasks/ci/licenses.rake
```diff
@@ -0,0 +1,23 @@
...
+      STDOUT.puts 'Checking license compliance'
+      unless system("bundle exec bin/license_finder")
+        STDERR.puts "*** License compliance test failed  ***"
+        exit 1
+      end
+    end
+    desc 'Generates a report with the dependencies and their licenses'
+    task :report do
+      STDOUT.puts 'Generating report...'
+      exec("bundle exec bin/license_finder report --format=xml > #{Rails.root.join('doc/licenses/licenses.xml')}")
```
mikz: This can be done by the RUBYOPT instead of having a custom binary, right @ddcesare ?

Then the whole rake and binary could be gone.

It is not pretty, but works. So I'm ok with it, but not happy :)

mikz:
@jamaga please add description to the PR.

lib/tasks/ci/licenses.rake
ddcesare: yes, it can be done by RUBYOPT

ddcesare:
@jamaga you can see in  https://github.com/3scale/zync/blob/master/doc/dependency_decisions.yml#L32-L39 that there are 2 groups ignored... that's why there are less in our report...

lib/tasks/ci/licenses.rake
ddcesare: Well, we are not using any custom binary here, just plain ol' license_finder gem

hallelujah: Not really as it uses bundler. And `RUBYOPT` takes over before LIBPATH resolution by bundler

.rubocop.yml
```diff
@@ -1158,4 +1158,4 @@ Style/WordArray:
   Enabled: false
 
 Style/FrozenStringLiteralComment:
-  Enabled: true
+  Enabled: false
```
ddcesare: @jamaga This was explicitly enabled before, did @mikz or @hallelujah agreed to change it?

jamaga: was to fix error in Codeclimate `Missing frozen string literal comment` @hallelujah should approve though

ddcesare: The way to fix it is actually freezing those strings, not disabling the linter unless it's not wanted anymore...

hallelujah: Not approved at all :)

lib/tasks/ci/licenses.rake
mikz: It would work if rake would be invoked by bundler then another exec would inherit the load paths and start just fine.

PR #46, opened 2017-08-17
Licenses xml

jamaga:
fixing codeclimate problem not by disabling feature but by adding 2 comments in files `# frozen_string_literal: true`

related: 
https://github.com/3scale/zync/pull/45
https://github.com/3scale/zync/pull/44


bin/license_finder
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+#!/usr/bin/env ruby
```
hallelujah: That should be the first line always

jamaga: OK

PR #69, opened 2017-10-02
expose job stats as prometheus metrics endpoint

mikz:
show information about how many jobs of each kind are scheduled

```
# TYPE scheduled_jobs job_stats
# HELP scheduled_jobs Que Jobs to be executed
scheduled_jobs{job="UpdateJob"} 5
# TYPE retried_jobs job_stats
# HELP retried_jobs Que Jobs to retried
retried_jobs{job="UpdateJob"} 3
# TYPE pending_jobs job_stats
# HELP pending_jobs Que Jobs that should be already running
pending_jobs{job="UpdateJob"} 2
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,6 +39,8 @@ gem 'message_bus' # for publishing notifications about integration status
 
 gem 'validate_url'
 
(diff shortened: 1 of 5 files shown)


jmprusi:
Thanks ! <3

PR #72, opened 2017-10-05
[Documentation] Adding a CONTRIBUTING.md file

hallelujah:
Part of https://issues.jboss.org/browse/THREESCALE-285

Simple contributing.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@ Regarding code style like indentation and whitespace, **follow the conventions y
 You can look at `.rubocop.yml` and `.codeclimate.yml`
 
 ## Modifying the code
-First, ensure that you have installed Ruby 2.2+ and Postgresql
+First, ensure that you have installed Ruby 2.3+ and recent PostgreSQL.
 
 1. Fork and clone the repo.
 1. Run `bundle install` to install all dependencies.
@@ -17,6 +17,5 @@ Assuming that you don't see any red, you're ready to go.
 1. Create a new branch, please don't work in your `master` branch directly.
 1. Add failing tests for the change you want to make. Run `rails test` to see the tests fail.
 1. Fix stuff.
-1. Run `rails test` to see if the tests pass. Repeat steps 2-4 until done.
-1. Update the documentation and Changelog to reflect any changes.
+1. Run `rails test` to see if the tests pass. Repeat steps 2-3 until done.
 1. Push to your fork and submit a pull request.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,22 @@
...
+
+Assuming that you don't see any red, you're ready to go.
+
+## Submitting pull requests
+
+1. Create a new branch, please don't work in your `master` branch directly.
+1. Add failing tests for the change you want to make. Run `rails test` to see the tests fail.
+1. Fix stuff.
+1. Run `rails test` to see if the tests pass. Repeat steps 2-4 until done.
+1. Update the documentation and Changelog to reflect any changes.
```
mikz: There is no documentation or changelog :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,22 @@
+# Contributing
+
+### Code style
+Regarding code style like indentation and whitespace, **follow the conventions you see used in the source already.**
+You can look at `.rubocop.yml` and `.codeclimate.yml`
+
+## Modifying the code
+First, ensure that you have installed Ruby 2.2+ and Postgresql
```
mikz: Ruby > 2.3. We probably should lock that in a gemfile.

CONTRIBUTING.md
andrewdavidmackenzie: that's the first update! :-)

CONTRIBUTING.md
andrewdavidmackenzie: Postgres min version? 9.5?

CONTRIBUTING.md
```diff
@@ -0,0 +1,22 @@
...
+1. Run `bundle install` to install all dependencies.
+
+Assuming that you don't see any red, you're ready to go.
+
+## Submitting pull requests
+
+1. Create a new branch, please don't work in your `master` branch directly.
+1. Add failing tests for the change you want to make. Run `rails test` to see the tests fail.
+1. Fix stuff.
+1. Run `rails test` to see if the tests pass. Repeat steps 2-4 until done.
```
andrewdavidmackenzie: steps 2-3 :-)    or it's infinite a loop within a loop :-)

CONTRIBUTING.md
andrewdavidmackenzie: Maybe something about issues? e.g. look for easy issues to fix? or submit issues for problems you have using the supplied template...?

mikz: I think it is fine like this. If there are easy to fix issues we will label them as such and it should be self-describing.

CONTRIBUTING.md
mikz: Not sure about the Postgres version. So better keep it like this.

PR #78, opened 2017-11-16
[prometheus] improve prometheus metrics

mikz:
more metrics about processed jobs
those new  are runtime only, so will have to be aggregated by prometheus
the previous ones were from database, so all nodes would give same results

```
# TYPE scheduled_jobs gauge
# HELP scheduled_jobs Que Jobs to be executed
scheduled_jobs{job="UpdateJob"} 4
# TYPE retried_jobs gauge
# HELP retried_jobs Que Jobs to retried
retried_jobs{job="UpdateJob"} 4
# TYPE pending_jobs gauge
# HELP pending_jobs Que Jobs that should be already running
# TYPE job_retries counter
# HELP job_retries A number of Jobs retried
job_retries{adapter="Que",job_name="UpdateJob"} 6.0
# TYPE failed_jobs counter
# HELP failed_jobs A number of Jobs errored
failed_jobs{adapter="Que",job_name="UpdateJob"} 6.0
# TYPE performed_jobs counter
# HELP performed_jobs A number of Jobs performed
performed_jobs{adapter="Que",job_name="UpdateJob"} 4.0
performed_jobs{adapter="Que",job_name="ProcessEntryJob"} 4.0
# TYPE enqueued_jobs counter
# HELP enqueued_jobs A number of Jobs enqueued
enqueued_jobs{adapter="Que",job_name="UpdateJob"} 10.0
enqueued_jobs{adapter="Que",job_name="ProcessEntryJob"} 4.0
# TYPE jobs_histogram histogram
# HELP jobs_histogram A histogram of Jobs perform times
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="0.005"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="0.01"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="0.025"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="0.05"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="0.1"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="0.25"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="0.5"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="1"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="2.5"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="5"} 1.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="10"} 1.0
jobs_histogram_bucket{adapter="Que",job_name="UpdateJob",le="+Inf"} 10.0
jobs_histogram_sum{adapter="Que",job_name="UpdateJob"} 4626.851000000001
jobs_histogram_count{adapter="Que",job_name="UpdateJob"} 10.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="0.005"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="0.01"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="0.025"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="0.05"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="0.1"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="0.25"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="0.5"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="1"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="2.5"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="5"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="10"} 0.0
jobs_histogram_bucket{adapter="Que",job_name="ProcessEntryJob",le="+Inf"} 4.0
jobs_histogram_sum{adapter="Que",job_name="ProcessEntryJob"} 1263.579
jobs_histogram_count{adapter="Que",job_name="ProcessEntryJob"} 4.0
# TYPE jobs_summary summary
# HELP jobs_summary A summary of Jobs perform times
jobs_summary{adapter="Que",job_name="UpdateJob",quantile="0.5"} 113.348
jobs_summary{adapter="Que",job_name="UpdateJob",quantile="0.9"} 171.259
jobs_summary{adapter="Que",job_name="UpdateJob",quantile="0.99"} 171.259
jobs_summary_sum{adapter="Que",job_name="UpdateJob"} 4626.851000000001
jobs_summary_count{adapter="Que",job_name="UpdateJob"} 10
jobs_summary{adapter="Que",job_name="ProcessEntryJob",quantile="0.5"} 18.095
jobs_summary{adapter="Que",job_name="ProcessEntryJob",quantile="0.9"} 46.349000000000004
jobs_summary{adapter="Que",job_name="ProcessEntryJob",quantile="0.99"} 46.349000000000004
jobs_summary_sum{adapter="Que",job_name="ProcessEntryJob"} 1263.579
jobs_summary_count{adapter="Que",job_name="ProcessEntryJob"} 4
```

### DIFF
--- a/config/initializers/prometheus.rb
+++ b/config/initializers/prometheus.rb
@@ -15,3 +15,5 @@ prometheus.register(retried_job_stats)
 pending_job_stats = Prometheus::JobStats.new(:pending_jobs, 'Que Jobs that should be already running')
 pending_job_stats.filter('run_at < now()')
 prometheus.register(pending_job_stats)
+
+require 'prometheus/active_job_subscriber'
--- a/lib/prometheus/active_job_subscriber.rb
+++ b/lib/prometheus/active_job_subscriber.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+module Prometheus
+
+  ## ActiveJob Subscriber to record Prometheus metrics.
+  ## Those metrics are per process, so they have to be aggregated by Prometheus.
+  class ActiveJobSubscriber < ActiveSupport::Subscriber
+
+    prometheus = Prometheus::Client.registry or raise 'Missing Prometheus client registry'
+
+    metrics = {
+        retried_jobs: prometheus.counter(:job_retries, 'A number of Jobs retried'),
+        failed_jobs: prometheus.counter(:failed_jobs, 'A number of Jobs errored'),
+        performed_jobs: prometheus.counter(:performed_jobs, 'A number of Jobs performed'),
+        enqueued_jobs: prometheus.counter(:enqueued_jobs, 'A number of Jobs enqueued'),
+        histogram: prometheus.histogram(:jobs_histogram, 'A histogram of Jobs perform times'),
+        summary: prometheus.summary(:jobs_summary, 'A summary of Jobs perform times'),
+    }
+    METRICS = OpenStruct.new(metrics).freeze
+
+    def initialize
+      super
+      @metrics = METRICS
+    end
+
+    def enqueue(event)
+      payload = event.payload
+      labels = extract_labels(payload)
+
+      if payload.fetch(:job).executions == 0
+        enqueued_jobs.increment(labels)
+      else
+        retried_jobs.increment(labels)
+      end
+    end
+
+    alias enqueue_at enqueue
+
+    def perform(event)
+      payload = event.payload
+      labels = extract_labels(payload)
+
+      observe_perform(payload, labels)
+      observe_duration(event, labels)
+    end
+
+    private
+
+    def observe_duration(event, labels)
+      duration = event.duration
+
+      histogram.observe(labels, duration)
+      summary.observe(labels, duration)
+    end
+
+    def observe_perform(payload, labels)
+      ex = payload[:exception_object]
+
+      if ex
+        # not measuring failed job duration, but we could
+        failed_jobs.increment(labels)
+      else
+        performed_jobs.increment(labels)
+      end
+    end
+
+    def extract_labels(payload)
+      {
+          adapter: payload.fetch(:adapter).class.name.demodulize.remove('Adapter'),
+          job_name: payload.fetch(:job).class.name
+      }
+    end
+
+    delegate :performed_jobs, :enqueued_jobs, :retried_jobs, :histogram, :summary, :failed_jobs,
+             to: :@metrics
+
+    attach_to :active_job
+  end
+end


orimarti:
Some things to have into account @mikz:

- Would be good to have some prefix like zync_ on all the metrics, as it's more readable when we have all metrics mixed with all the other ones on prometheus.
- Would be good to have the metrics on another port not directly on the service, as it will be visible to all internet, and don't think it's very nice...
- What runtime exactly means? At minimum you should be getting all the metrics for 5 or 10 seconds which could be the interval that prometheus takes the metrics.

orimarti:
Don't understand this graph:
https://prometheus.3sca.net/dashboard/db/zync?orgId=1&from=now-4d&to=now&panelId=4&fullscreen&edit
Looks like ProcessIntegrationJob is acumulative and the other ones are going down after some time.
Is it correct @mikz?

mikz:
@orimarti 

1. Ok. Should I rename all the metrics?
2. Yes, but not necessary now as we do not need to expose this service publicly.
3. Runtime means that each process has own metrics - different numbers. That means if we run puma in clustered mode with > 1 processes you'll not get all stats from all processes. It also means each Pod on OpenShift will provide different numbers. Those previous metrics are the same from all processes because they are taken from database. Any suggestions how to name them to highlight this distinction?

orimarti:
> Ok. Should I rename all the metrics?

Yes, would be better, maybe out of this PR.

> Runtime means that each process has own metrics - different numbers. That means if we run puma in clustered mode with > 1 processes you'll not get all stats from all processes. It also means each Pod on OpenShift will provide different numbers. Those previous metrics are the same from all processes because they are taken from database. Any suggestions how to name them to highlight this distinction?

We can treat the case when each Pod has one puma, but if we have more than one process, don't see an easy way to take data from each process, so here my recomendation would be to run one process on puma. We have to take into account that prometheus takes data every X seconds, so probably we should use push-gateway if we don't want to lose the metrics that are not taken on the moment that prometheus is doing the scrape.

mikz:
@orimarti so nothing to be done for this PR?

orimarti:
> @orimarti so nothing to be done for this PR?

Well, IMO we should change it to push gateway sending the data through the network, or use the database if it's posible the same we were doing on the old metrics.

mikz:
@orimarti I'm happy to integrate the push gateway. These metrics are not in database, that would be my first choice.

mikz:
@orimarti when using Pushgateway zync would have to push to it only every N seconds. Having internal timer to update it like prometheus scraper would. Because it does not do any aggregations.
So you'd still lose data if something crashes. https://github.com/prometheus/pushgateway#non-goals

Not sure if Pushgateway is the right fit here.

orimarti:
We'll have to use statsd probably, let me investigate little bit more what are the options here.

mikz:
This is pretty much like every other metric gathered from other apps is going to work. 
I think we have to accept that if the app crashes the data between the scrape interval are lost.
That is fundamental property of prometheus.

I'm merging this. If there is a proposal to move to something else, then we can redo this.

PR #82, opened 2018-02-21
update httpclient to support custom certificate stores [THREESCALE-466]

mikz:
fixes [THREESCALE-466](https://issues.jboss.org/browse/THREESCALE-466)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,9 @@ gem 'que'
 gem 'bugsnag'
 gem 'bugsnag-capistrano', require: false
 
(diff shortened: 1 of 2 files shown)


mayorova:
@hallelujah The other gem was [OAuth2::Client](https://github.com/oauth-xx/oauth2), I think it's based on Faraday, that respects `SSL_CERT_FILE`. I guess we just need QA to confirm it all works :)

mikz:
OAuth2::Client is configured to use `httpclient` so it works as well.

PR #83, opened 2018-02-21
[fetch] detect when running on openshift and use services instead of routes

mikz:
If there is a way to reach System internally via Service it should be preferred.

### DIFF
--- a/app/adapters/three_scale/api/instrumented_http_client.rb
+++ b/app/adapters/three_scale/api/instrumented_http_client.rb
@@ -5,6 +5,20 @@ require 'securerandom'
 # Custom HTTP Client for 3scale API client with added instrumentation.,
 class ThreeScale::API::InstrumentedHttpClient < ThreeScale::API::HttpClient
 
+  def initialize(**)
+    super
+
+    if (system_provider_port = ENV['SYSTEM_PROVIDER_PORT'].presence)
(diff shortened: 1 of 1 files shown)


app/adapters/three_scale/api/instrumented_http_client.rb
```diff
@@ -5,6 +5,21 @@
 # Custom HTTP Client for 3scale API client with added instrumentation.,
 class ThreeScale::API::InstrumentedHttpClient < ThreeScale::API::HttpClient
 
+  def initialize(**)
+    super
+
+    if (system_provider_port = ENV['SYSTEM_PROVIDER_PORT'].presence)
```
mikz: @jmprusi I think this is pretty decent way of overriding endpoints with local services. Instead of going through a Route use the internal Service and set `X-Forwarded-Host` and `X-Forwarded-Proto`. 
This could help us get rid of the `BACKEND_OVERRIDE` in APIcast and it could just detect it if proper ENV is set.

hallelujah: `SYSTEM_PROVIDER_PORT` is an env automatically set by Openshift right?

mikz: Yes.

PR #88, opened 2018-07-03
Send zync token to authenticate to System

hallelujah:
Mutual authentication with system.
It uses the same token that System workers to authenticate with Zync.
Zync will query system with this token in the header to identify itself

Ideas could be to use BasicAuth but we would need to modify System code base.
Can be improved later I think

### DIFF
--- a/app/adapters/three_scale/api/instrumented_http_client.rb
+++ b/app/adapters/three_scale/api/instrumented_http_client.rb
@@ -58,10 +58,10 @@ class ThreeScale::API::InstrumentedHttpClient < ThreeScale::API::HttpClient
   protected
 
   def headers
-    super.merge(extra_headers)
+    super.merge(self.class.extra_headers)
   end
 
-  def extra_headers
+  def self.extra_headers
     { 'X-Request-Id' => SecureRandom.uuid, 'X-Zync-Token' => Rails.application.secrets.authentication[:token] }.compact
   end
 
--- a/app/adapters/three_scale/api/instrumented_http_client.rb
+++ b/app/adapters/three_scale/api/instrumented_http_client.rb
@@ -58,7 +58,11 @@ class ThreeScale::API::InstrumentedHttpClient < ThreeScale::API::HttpClient
   protected
 
   def headers
-    super.merge('X-Request-Id' => SecureRandom.uuid)
+    super.merge(extra_headers)
+  end
+
+  def extra_headers
+    { 'X-Request-Id' => SecureRandom.uuid, 'X-Zync-Token' => Rails.application.secrets.authentication[:token] }.compact
   end
 
   def build_request(type, path, params)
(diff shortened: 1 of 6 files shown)


app/adapters/three_scale/api/instrumented_http_client.rb
```diff
@@ -57,8 +58,13 @@ def parse(response)
 
   protected
 
+  def zync_token
+    authentication = Rails.application.secrets.authentication || {}.freeze
```
Martouta: `Rails.application.secrets.authentication` is currently empty 😞

app/adapters/three_scale/api/instrumented_http_client.rb
```diff
@@ -57,8 +58,13 @@ def parse(response)
...
+    authentication.fetch(:token) { return }
+  end
+
   def headers
-    super.merge('X-Request-Id' => SecureRandom.uuid)
+    super.merge('X-Request-Id' => SecureRandom.uuid, 'X-Zync-Token' => zync_token)
```
Martouta: There is a bug here. When `zync_token` returns nil, we are sending this header anyway and `build_request` does inside a `headers.each do |key, value| ... value.strip ... end` where that value is expected to be an String.
We can fix it by either returning `''` instead of nil or by not sending this header at all when it is empty, which is what makes more sense to me but both seem good to me.

app/adapters/three_scale/api/instrumented_http_client.rb
```diff
@@ -58,13 +58,12 @@ def parse(response)
...
-  def zync_token
-    authentication = Rails.application.secrets.authentication || {}.freeze
-    authentication.fetch(:token) { return }
-  end
 
   def headers
-    super.merge('X-Request-Id' => SecureRandom.uuid, 'X-Zync-Token' => zync_token)
+    super.merge(
+      'X-Request-Id' => SecureRandom.uuid,
+      'X-Zync-Token' => Rails.application.secrets.authentication.fetch(:token) { return }
```
Martouta: This is going to raise a "NoMethodError: 'strip' is not defined for nil" when ENV['ZYNC_AUTHENTICATION_TOKEN'] is not defined.

app/adapters/three_scale/api/instrumented_http_client.rb
```diff
@@ -58,7 +58,10 @@ def parse(response)
   protected
 
   def headers
-    super.merge('X-Request-Id' => SecureRandom.uuid)
+    super.merge('X-Request-Id' => SecureRandom.uuid).tap do |headers|
```
mikz:
```ruby
def headers
  super.merge(extra_headers)
end

def extra_headers
  { 'X-Request-Id' => SecureRandom.uuid, 'X-Zync-Token' => Rails.application.secrets.authentication[:token] }.compact
end
```

`Hash#compact` returns a hash with non `nil` values.

PR #100, opened 2018-10-18
[keycloak] ability to configure extra client attributes

mikz:
Define yaml config with extra attributes sent to keycloak. This allows anyone to pass own yaml config and configure extra payload sent to Keycloak.

closes #91
closes #89


mayorova:
Nice :)

PR #111, opened 2018-10-19
Synchronize changes with the same client_id on the same service [THREESCALE-1346]

mikz:
Fixes [THREESCALE-1346](https://issues.jboss.org/browse/THREESCALE-1346).

By introducing another model Client and using System API to find the last App with specific client_id we can synchronize changes even when Application is deleted and another one created with the same app_id.

### DIFF
--- a/app/services/integration/keycloak_service.rb
+++ b/app/services/integration/keycloak_service.rb
@@ -12,23 +12,47 @@ class Integration::KeycloakService
   def call(entry)
     case entry.record
     when Proxy then handle_test
-    when Application then handle_application(entry)
+    when Application then ClientFromApplication.call(entry)
+    when Client then handle_client(entry)
     else handle_rest(entry)
     end
   end
 
-  def handle_application(entry)
+  def handle_client(entry)
     client = build_client(entry)
-
-    unless client.id
-      # Not OAuth
-      return
-    end
 
     if persist?(client)
       persist(client)
     else
       remove(client)
+    end
+  end
+
+  # Convert Application to Client and trigger new update from the API.
+  # Creates new Client if needed and triggers UpdateJob for it.
+  class ClientFromApplication
+    def self.call(entry)
+      new(entry).call
+    end
+
+    attr_reader :tenant, :client_id, :scope
+
+    def initialize(entry)
+      @client_id = entry.data&.dig('client_id')
+      @tenant = entry.tenant
+      @scope = Client.for_service(entry.record.service)
+    end
+
+    def call
+      return unless client_id
+
+      model = Model.create_record!(tenant) do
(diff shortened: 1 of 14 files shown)


app/services/integration/keycloak_service.rb
```diff
@@ -12,26 +12,50 @@ def initialize(integration)
...
     else
       remove(client)
     end
   end
 
+  # Convert Application to Client and trigger new update from the API.
+  # Creates new Client if needed and triggers UpdateJob for it.
+  class ClientFromApplication
+    def self.call(entry)
+      new(entry).call
```
guicassolato: I've found myself doing the same more than once recently. It would be nice to have a more DSL-ish way to do it, like:
```ruby
define_singleton_method :call
```
P.S. I know `Object#define_singleton_method` is not supposed to be used like that, but if not that, some other name.

hallelujah:
Looks good.

PS: I wanted to check but I can't find the part of the code where it synced back to Multi Tenant when something changes in Keycloak (like the secret for example).

mikz:
@hallelujah because it is not synced back :) It is one-way synchronization.

app/services/integration/keycloak_service.rb
mikz: I'm not sure using `define_singleton_method` is such a win.

```ruby
def self.call(entry)
  new(entry).call
end
```

vs

```
define_singleton_method do |entry|
  new(entry).call
end
```

If I'd find more of these I'd probably go for module:

```ruby
module ServiceClass
  def call(*args, **opts, &block)
    new(*args, **opts, &block).call
  end
end

class MyService
  include ServiceClass 

  def new(foo)
  end

  def call
  end
```

guicassolato: _Off topic –_ I've found myself doing the same more than once recently. It would be nice to have a more DSL-ish way to do it, like:
```ruby
define_singleton_sender_for :call
```

```ruby
class Object
  def define_singleton_sender_for(*instance_method_symbols)
    instance_method_symbols.each do |method_symbol|
      define_singleton_method(method_symbol) { |*args| new(*args).send(method_symbol) }
    end
  end
end
```

guicassolato: I also had thought about
```ruby
define_singleton_sender_for :call
```

```ruby
class Object
  def define_singleton_sender_for(*instance_method_symbols)
    instance_method_symbols.each do |method_symbol|
      define_singleton_method(method_symbol) { |*args| new(*args).send(method_symbol) }
    end
  end
end
```

But maybe you're right. That can easily lead to violation of Single Responsibility, unless you use it only for `call` and `call!` which is indeed not a big win.

PR #114, opened 2018-10-26
Add zync openshift manifests

slopezz:
Added zync openshift manifests. 

It has been also added a Makefile to simplify the openshift application creation.

### DIFF
--- a/openshift/03-buildconfig-template.yml
+++ b/openshift/03-buildconfig-template.yml
@@ -1,0 +1,46 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: zync-buildconfig
+objects:
+- apiVersion: v1
+  kind: BuildConfig
+  metadata:
+    labels:
+      app: zync
+    name: zync
+  spec:
+    output:
+      to:
+        kind: ImageStreamTag
+        name: zync:latest
+    postCommit:
+      command:
+      - /bin/bash
+      - -c
+      - bin/rails runner 'puts Rails.env'
+    runPolicy: Serial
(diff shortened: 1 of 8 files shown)


openshift/05-services-template.yml
```diff
@@ -0,0 +1,39 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: zync-services
+objects:
+- apiVersion: v1
+  kind: Service
+  metadata:
+    annotations:
+      service.alpha.openshift.io/dependencies: '[{"name": "zync-database", "kind":
```
mikz: This can be removed, has been removed in OpenShift.

openshift/03-buildconfig-template.yml
```diff
@@ -0,0 +1,41 @@
...
+        from:
+          kind: ImageStreamTag
+          name: ruby-24-centos7:latest
+      type: Source
+    triggers:
+    - github:
+        secret: "${GITHUB_SECRET}"
+      type: GitHub
+    - type: ImageChange
+    - type: ConfigChange
```
maneta: I Think it would be good to have a postCommit like we have in https://github.com/3scale/apicast-cloud-hosted/blob/master/openshift/03-build-config.yml#L10-L15 

Maybe @mikz could point us to a good postCommit

mikz:
@slopezz I don't think post-commit hooks are a good idea for enforcing workflow/lint. For that, there is a CI. post-commit hooks are for *you* to enforce *your* flow, not the project.

PR #146, opened 2019-02-06
[DO NOT MERGE] Can create and delete apps

Martouta:
This project has 2 bugs in master at the moment that I am creating this PR when integrating with Porta/System and Keycloak/RHSSO:
- It can not create applications in Keycloak when created in Porta
- It can not delete applications in Keycloak when deleted in Porta

This PR is NOT meant to be merged and it is NOT real fixes for those problems. But It works and that is enough for just testing around the integration for the first time now that people from the system team will check this project for the first time.

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -2,4 +2,5 @@
 
 class Service < ApplicationRecord
   belongs_to :tenant
+  has_many :proxies
 end
--- a/app/services/integration/keycloak_service.rb
+++ b/app/services/integration/keycloak_service.rb
@@ -38,7 +38,7 @@ class Integration::KeycloakService
     attr_reader :tenant, :client_id, :scope
 
     def initialize(entry)
-      @client_id = entry.data&.dig('client_id')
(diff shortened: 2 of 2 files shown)


mikz:
@Martouta so, are you saying it does not do the only thing it is supposed to do?

Martouta:
@mikz exactly 😟 in the case of creating I don't know if it is always, but in the case of deleting it is always.
They both have Jira issues and are explained there:
https://issues.jboss.org/browse/THREESCALE-1480
https://issues.jboss.org/browse/THREESCALE-1791

Martouta:
Comment from Michal:

> It must be some edge case, because it passes QE tests (at least passed at the release time)
> We would catch if it would not work at all
> Ping me if you need help how to write the test to cover it
> https://github.com/3scale/zync/pull/111 had some complicated tests
> https://github.com/3scale/zync/pull/109

Martouta:
This was not meant to be merged but just a clue for the team and it is not needed anymore 😄

PR #156, opened 2019-03-07
OAuth allowed flows per application

mikz:
Implement OAuth allowed flows for each Application based on information in https://issues.jboss.org/browse/THREESCALE-1948

Part of https://issues.jboss.org/browse/THREESCALE-774


mikz:
Verified it works with latest changes in porta.

PR #164, opened 2019-03-19
THREESCALE-1791 - 404 Application cannot be deleted on Keycloak

macejmic:
Fixes [THREESCALE-1791](https://issues.jboss.org/browse/THREESCALE-1791)

### DIFF
--- a/app/services/integration/keycloak_service.rb
+++ b/app/services/integration/keycloak_service.rb
@@ -38,7 +38,7 @@ class Integration::KeycloakService
     attr_reader :tenant, :client_id, :scope
 
     def initialize(entry)
-      @client_id = entry.data&.dig('client_id')
+      @client_id = entry.last_known_data.dig('client_id')
       @tenant = entry.tenant
       @scope = Client.for_service(entry.record.service)
     end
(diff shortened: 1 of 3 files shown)


app/services/integration/keycloak_service.rb
```diff
@@ -38,7 +38,7 @@ def self.call(entry)
     attr_reader :tenant, :client_id, :scope
 
     def initialize(entry)
-      @client_id = entry.data&.dig('client_id')
+      @client_id = (entry.data || entry.previous_data)&.dig('client_id')
```
Martouta: We can fix the nil check 😄 
```suggestion
      @client_id = (entry.data || entry.previous_data || {}).dig('client_id')
```

Martouta: And maybe we should even add a method in `entry` for this 😄 
image

Maybe something like:
```ruby
class Entry
  def last_known_client_id
    (data || previous_data || {}).dig('client_id')
  end
end
```
And here: `entry.last_known_client_id`

mikz: Entry is generic for any type as zync can sync more than just Applications. So Entry should not really know about `client_id` (but surely can know about previous data).

Martouta: Ok, then maybe:
```ruby
class Entry
  def last_known_data
    (data || previous_data || {})
  end
end
```
and then here:
`entry.last_known_data.dig('client_id')`

PR #175, opened 2019-04-10
Que upgrade

mikz:
* Upgrade que to 1.0.0beta3 (using pubsub in pg)
* Listen on a second port (9393) and respond with prometheus metrics
* Do not queue requests into puma

### DIFF
--- a/openshift/04-deploymentconfig-template.yml
+++ b/openshift/04-deploymentconfig-template.yml
@@ -25,7 +25,7 @@ objects:
     template:
       metadata:
         annotations:
-          prometheus.io/port: "8080"
+          prometheus.io/port: "9393"
           prometheus.io/scrape: "true"
         labels:
           app: zync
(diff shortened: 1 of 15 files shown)


openshift/04-deploymentconfig-template.yml
```diff
@@ -98,6 +100,80 @@ objects:
...
+          args:
+          - que
+          - --worker-count
+          - 10
+          image: zync:latest
+          imagePullPolicy: Always
+          livenessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /status/live
```
mikz: nope, this is not listening on any port

openshift/04-deploymentconfig-template.yml
```diff
@@ -98,6 +100,80 @@ objects:
...
+              port: 8080
+              scheme: HTTP
+            initialDelaySeconds: 10
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 60
+          name: metrics
```
mikz: que or background workers or something, not metrics

openshift/04-deploymentconfig-template.yml
```diff
@@ -98,6 +100,80 @@ objects:
...
+          ports:
+          - name: zync
+            containerPort: 8080
+            protocol: TCP
+          - name: metrics
```
mikz: this is exposed on the original container

PR #176, opened 2019-04-10
[metrics] Manage metrics in a separated container

maneta:
This allows prometheus to scrape the metrics from a different
container and port.

### DIFF
--- a/openshift/04-deploymentconfig-template.yml
+++ b/openshift/04-deploymentconfig-template.yml
@@ -25,7 +25,7 @@ objects:
     template:
       metadata:
         annotations:
-          prometheus.io/port: "8080"
+          prometheus.io/port: "9393"
           prometheus.io/scrape: "true"
         labels:
           app: zync
@@ -62,6 +62,8 @@ objects:
               secretKeyRef:
                 key: ZYNC_AUTHENTICATION_TOKEN
                 name: zync
+          - name: RAILS_MAX_THREADS
+            value: '10'
           - name: RAILS_LOG_LEVEL
             value: info
           image: zync:latest
@@ -78,7 +80,11 @@ objects:
             timeoutSeconds: 60
           name: zync
           ports:
-          - containerPort: 8080
+          - name: zync
+            containerPort: 8080
+            protocol: TCP
+          - name: metrics
+            containerPort: 9393
             protocol: TCP
           readinessProbe:
             failureThreshold: 3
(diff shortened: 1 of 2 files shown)


openshift/04-deploymentconfig-template.yml
```diff
@@ -98,6 +100,80 @@ objects:
...
+                name: zync
+          - name: RAILS_LOG_LEVEL
+            value: info
+          args:
+          - que
+          - --worker-count
+          - 10
+          image: zync:latest
+          imagePullPolicy: Always
+          livenessProbe:
```
mikz: que is the background work processor, it does not start listening on any port

openshift/04-deploymentconfig-template.yml
```diff
@@ -98,6 +100,80 @@ objects:
...
+            failureThreshold: 3
+            httpGet:
+              path: /status/live
+              port: 8080
+              scheme: HTTP
+            initialDelaySeconds: 10
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 60
+          name: metrics
```
mikz: que is processing jobs in background, not exposing metrics

openshift/04-deploymentconfig-template.yml
```diff
@@ -98,6 +100,80 @@ objects:
...
+          ports:
```
mikz: no ports here

openshift/04-deploymentconfig-template.yml
maneta: Ok Do you think we need some readiness/liveness for this container? If you think so, which command you think would be good for that?

openshift/04-deploymentconfig-template.yml
maneta: done on 031e3fe1281af2f2942a74e57e6aa37fb372f302

openshift/04-deploymentconfig-template.yml
maneta: done on 031e3fe1281af2f2942a74e57e6aa37fb372f302

maneta: done on 031e3fe1281af2f2942a74e57e6aa37fb372f302

openshift/04-deploymentconfig-template.yml
mikz: I think it is not needed. if the process dies it will kill the container and there is no readiness really. it is all background processing.

maneta: ok Cool

PR #182, opened 2019-04-11
APICAST_REGISTRY_URL no longer required for porta to run  [THREESCALE-2060]

josemigallas:
JIRA: [[THREESCALE-2060] We shouldn't need to have a valid json in apicast_registry_url in the config to run porta locally](https://issues.jboss.org/browse/THREESCALE-2060)

### DIFF
--- a/INTEGRATE.md
+++ b/INTEGRATE.md
@@ -20,7 +20,6 @@ Go to the folder where you have zync installed and run it in the port 5000.
 5. Run Porta
 These configurations are needed:
   - In `config/rolling_updates.yml`, make sure to have `apicast_v2: true` and `apicast_oidc: true`.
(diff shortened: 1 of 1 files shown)


Martouta:
@josemigallas This shouldn't have been merged without merging https://github.com/3scale/porta/pull/736 first.
For this time it doesn't matter because we plan to merge it soon and as it is documentation here and not real code, nothing will break and the maximum that can happen is that the integration doesn't work for someone who follows the instructions these days until the other PR is merged.
But for the next time it would be nice to have a `[WIP]` on the title to prevent merging a PR that depends on another not merged PR 😄

mikz:
Or write it in the description.

PR #196, opened 2019-05-31
Generic HTTP adapter

mikz:
An adapter implementing https://github.com/3scale/zync/commit/1228d796316fb8d8a7c32e6171e8997d23a465f3 protocol.
Depends on https://github.com/3scale/porta/pull/829


app/adapters/abstract_adapter.rb
```diff
@@ -0,0 +1,294 @@
...
+
+  def authentication=(value)
+    oidc.access_token = value
+  end
+
+  def authentication
+    oidc.access_token.token
+  end
+
+  def create_client(_)
```
hallelujah: Those methods have only one argument?
Pinpointing: Better use `create_client(_client)`

mikz: Yeah, I did those by hand. Rubocop autocorrect would do correctly `_client`.

app/adapters/abstract_adapter.rb
```diff
@@ -0,0 +1,294 @@
...
+    raise NotImplementedError, __method__
+  end
+
+  def read_client(_)
+    raise NotImplementedError, __method__
```
hallelujah: Not fan of `NotImplementedError` because it is not really an OS implementation. Ancestor is `ScriptError` and not `StandardError` also.

Preferably `NoMethodError`

```ruby
2.3.8 (main):0 > NotImplementedError.ancestors
=> [NotImplementedError,
 ScriptError,
 Exception,
 ActiveSupport::Dependencies::Blamable,
 Object,
 ERB::Util,
 PP::ObjectMixin,
 JSON::Ext::Generator::GeneratorMethods::Object,
 ActiveSupport::Dependencies::Loadable,
 Kernel,
 BasicObject]
```

mikz: Good point!

app/adapters/abstract_adapter.rb
```diff
@@ -0,0 +1,294 @@
...
+      end
+    end
+  end
+
+  def parse(response)
+    body = self.class.parse_response(response)
+
+    raise InvalidResponseError, { response: response, message: body } unless response.ok?
+
+    params = body.try(:to_h) or return # no need to create client if there are no attributes
```
hallelujah: If the content-type is NULL_TYPE, `body` will be a `String`
https://github.com/3scale/zync/pull/196/files#diff-cbd4720b5437174ca7b9bcc0a46568adR99

mikz: Which is fine, because it will return `nil` and return.

app/adapters/generic_adapter.rb
```diff
@@ -0,0 +1,140 @@
...
+          **self.class.attributes,
+      }
+    end
+
+    def redirect_url=(val)
+      self.redirect_uris = [ val ].compact
+    end
+
+    def oidc_configuration=(config)
+      self.grant_types = GrantTypes.new(config)
```
hallelujah: Wouldn't it safer/easier to assign an Array here as you already define `to_ary` in https://github.com/3scale/zync/pull/196/files#diff-9ef24b12100d60cc4d037c50e0990ae0R95?

So you do not need to call `to_ary` in https://github.com/3scale/zync/pull/196/files#diff-9ef24b12100d60cc4d037c50e0990ae0R64

app/adapters/generic_adapter.rb
```diff
@@ -0,0 +1,140 @@
...
+      attr_reader :params
+    end
+    private_constant :GrantTypes
+  end
+
+  protected
+
+  def client_url(client_or_id)
+    id = client_or_id.to_param or raise 'missing client id'
+    (endpoint + "clients/#{id}").freeze
```
hallelujah:
```suggestion
    URI.join(endpoint, "clients/#{id}").freeze
```

app/adapters/generic_adapter.rb
mikz: I'll try to implement `as_json` instead of `to_ary`. I want to convert it at the latest moment possible.

app/services/discover_integration_service.rb
```diff
@@ -18,14 +18,16 @@ def call(integration)
...
             when Integration::Keycloak
               Integration::KeycloakService
+            when Integration::Generic
+              Integration::GenericService
             when integration
               Integration::EchoService
             else # the only one for now
               raise NotImplementedError
-            end
+            end.freeze
```
hallelujah: Is it expected to freeze classes at runtime? Can you explain me why? I do simply not get the use case

test/adapters/abstract_adapter_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    assert_kind_of URI, adapter.endpoint
+  end
+
+  test 'endpoint normalization' do
+    uri = URI('http://lvh.me:3000/auth/realm/name/')
+
+    assert_equal uri,
+                 subject.new('http://id:secret@lvh.me:3000/auth/realm/name').endpoint
+
+    assert_equal uri,
```
hallelujah:
```suggestion
```

hallelujah: Same test than the above :)

test/adapters/generic_adapter_test.rb
```diff
@@ -0,0 +1,172 @@
...
+  test 'timeout error' do
+    stub_request(:get, 'http://lvh.me:3000/auth/realm/name/.well-known/openid-configuration').
+        to_return(status: 200, headers: { 'Content-Type' => 'application/json' },
+                  body: { token_endpoint: 'protocol/openid-connect/token' }.to_json)
+
+    stub_request(:post, 'http://lvh.me:3000/auth/realm/name/protocol/openid-connect/token').to_timeout
+
+    adapter = GenericAdapter.new('http://id:secret@lvh.me:3000/auth/realm/name')
+
+    begin
```
hallelujah: Without begin/rescue

```ruby
error = assert_raise GenericAdapter::OIDC::AuthenticationError do
  adapter.test
end

assert_kind_of Faraday::TimeoutError, error.cause
assert error.bugsnag_meta_data.presence
```

app/services/discover_integration_service.rb
mikz: I wanted to freeze the instance. Well spotted.

test/adapters/generic_adapter_test.rb
mikz: I'll revert your d34e9d1. Feel free to apply those changes when they are working. But I need to merge & deploy this.

hallelujah: Fair enough, I will work on it

PR #198, opened 2019-06-04
Do not use begin/rescue in tests

hallelujah:
Stub Rails.logger.error

When fetching headers, we rescue any authentication error and set the
headers as application/json 
https://github.com/3scale/zync/blob/464cd87fda2275468ad4a94d9fbb54cfe98df10e/app/adapters/generic_adapter.rb#L136

The test rescues an error that is never raised

### DIFF
--- a/test/adapters/generic_adapter_test.rb
+++ b/test/adapters/generic_adapter_test.rb
@@ -35,16 +35,17 @@ class GenericAdapterTest < ActiveSupport::TestCase
         to_return(status: 200, headers: { 'Content-Type' => 'application/json' },
                   body: { token_endpoint: 'protocol/openid-connect/token' }.to_json)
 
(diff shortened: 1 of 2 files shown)


test/adapters/generic_adapter_test.rb
```diff
@@ -35,16 +35,29 @@ class GenericAdapterTest < ActiveSupport::TestCase
...
     adapter = GenericAdapter.new('http://id:secret@lvh.me:3000/auth/realm/name')
 
-    begin
+    log = Object.new
+    class << log
+      def error_object
+        @error
+      end
+
+      def error(object)
```
mikz: Why not using a MiniTest::Mock ?

akostadinov (dismissed):
I think this is a reasonable approach that can be improved another time.

PR #224, opened 2019-07-10
[k8s] ignore owner references that zync has no permission to

mikz:
When deployed by the operator, 3scale objects have the following owner reference:

```yaml
ownerReferences:
    - apiVersion: apps.3scale.net/v1alpha1
      kind: APIManager
      name: example-apimanager-s3
      uid: 3fcf9ead-a2f3-11e9-b57e-0a543109227c
      controller: true
      blockOwnerDeletion: true
```

And zync won't have permissions to read that object. So just stop the recursion and use the latest valid object as a root owner reference.

/cc @miguelsorianod

### DIFF
--- a/app/services/integration/kubernetes_service.rb
+++ b/app/services/integration/kubernetes_service.rb
@@ -61,6 +61,9 @@ class Integration::KubernetesService < Integration::ServiceBase
     end
 
     resource
+  rescue K8s::Error::Forbidden
+    # likely some resource like the operator
+    resource
   end
 
   def get_owner


miguelsorianod:
I understand this changes assumes that if there's an error reading the object it assumes it belongs to a Kind of object that it is a custom resource (owned by an operator).

By reviewing at the code it seems you are trying to get the DeploymentConfig name related to the pod where zync is deployed.

As we discussed I think that's not necessary at all. When deploying 3scale we are providing a label named "deploymentConfig" that contains the DeploymentConfig name and you could reconstruct the OwnerReference from that.
That label name and value is the label used as a selector by the DeploymentConfig, and is set in the PodTemplateSpec of the DeploymentConfig too so it should be reliable.

I also have observed that some controller (different than the 3scale-operator one) adds the 
"deploymentconfig" (notice lowercase c) and "deployment" labels into the pods (even if not specified in the podtemplatespec) so it seems there's also some kind of information added by some Kubernetes controller to relate the pod and the deploymentconfig it belongs.

In my opinion there's no need to recurse through OwnerReferences for that. A look for one of those labels could be performed and then the reconstruction of the OwnerReference could be created.

As we talked you think the recursion through OwnerReferences is more appropriate so I'm not going to oppose to this change either as you are the one working on this project and have deep knowledge of the code and the PR solution seems to work too as we tested in a cluster.

PR #257, opened 2019-09-30
Use UBI7 Ruby image

duduribeiro:
THREESCALE-3367
Support UBI7 (Universal Base Image) image for Zync

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,16 +4,18 @@ jobs:
     resource_class: small
     working_directory: /home/circleci/zync
     docker:
-      - image: circleci/postgres:10.9-alpine
-        environment:
-          POSTGRES_USER: postgres
-          POSTGRES_DB: zync
-    environment:
-      DATABASE_URL: postgresql://postgres:@127.0.0.1:5432/zync
+      - image: circleci/ruby:2.4.6
     steps:
+      - run:
+          name: Install Docker Compose
+          command: |
+            curl -L https://github.com/docker/compose/releases/download/1.19.0/docker-compose-`uname -s`-`uname -m` > ~/docker-compose
+            chmod +x ~/docker-compose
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -8,6 +8,11 @@ RUN source ${APP_ROOT}/etc/scl_enable \
 COPY . .
 ENV RAILS_LOG_TO_STDOUT=1
 USER root
+RUN rpm -ivh http://yum.postgresql.org/10/redhat/rhel-7-x86_64/pgdg-redhat-repo-42.0-5.noarch.rpm
```
hallelujah: Not really, we do not want to install other dependencies in this image

.circleci/config.yml
```diff
@@ -40,6 +49,15 @@ jobs:
           name: rails db:setup
           command: bundle exec bin/rails db:wait db:setup
 
+      - run:
+          name: rails db:setup for docker image
+          command: docker-compose run -e RAILS_ENV=test zync bin/rails db:drop db:create db:migrate
```
hallelujah: Avoid using docker compose when not necessary

Just build the image `docker build . --tag=$IMAGE_NAME`
Then use this image by `docker run $IMAGE_NAME`

It is just a smoke test not a full testing

.circleci/config.yml
```diff
@@ -11,6 +11,15 @@ jobs:
         DISABLE_SPRING: 1 # we can't really run spring as it hangs on local circleci build
         DATABASE_URL: postgres://postgres:@localhost/circle_test
     steps:
+      - run:
```
hallelujah: I would create an other entire job for that image building

.circleci/config.yml
```diff
@@ -40,6 +42,12 @@ jobs:
           name: rails db:setup
           command: bundle exec bin/rails db:wait db:setup
 
+      - run:
```
hallelujah: Please use another job

PR #265, opened 2019-10-22
Route termination options follow public url scheme

hallelujah:
If public route was created with `http:` scheme, do not enforce secure route edge redirect

Fixes THREESCALE-3545

#### TODO

- [x] Add tests for RouteSpec
- [x] Test in live OpenShift

#### Issues/Questions

The following does not prevent the merge of this PR but we should take into account that when the environment variable `KUBERNETES_ROUTE_TLS` is set to `1`, `t` or `true` then it will prevent creating 
non secure routes, which might be an issue as **Porta** has something different from **OpenShift**

This prevents removing the `:tls` option https://github.com/3scale/zync/blob/726962b98feb15f1859b0c1ad8ae72b7b6e71028/app/services/integration/kubernetes_service.rb#L6-L7

It is used in https://github.com/3scale/zync/blob/726962b98feb15f1859b0c1ad8ae72b7b6e71028/app/services/integration/kubernetes_service.rb#L263-L271

But the integration service builds all routes for both staging and production endpoint, so disabling the `maintain_tls_spec` for the integration service is not possible because for example, staging route can be HTTP and production route can be HTTPS

https://github.com/3scale/zync/blob/726962b98feb15f1859b0c1ad8ae72b7b6e71028/app/services/integration/kubernetes_service.rb#L153-L158

### Openshift testing

Using this Build config

```yaml
apiVersion: v1
kind: Template
metadata:
  name: "zync-build"
message: "3scale AMP zync builder"
objects:

- kind: "BuildConfig"
  apiVersion: "v1"
  metadata:
    name: "zync-build"
  spec:
    source:
      git:
        uri: "https://github.com/3scale/zync"
        ref: "THREESCALE-3545"
    output:
      to:
        kind: "ImageStreamTag"
        name: "amp-zync:latest"

    strategy:
      type: Docker
      dockerStrategy:
```

### Screenshots

image

image

### Caveats

Due to OpenShift router limitation,[ the HTTP and HTTPS ports are automatically set by OpenShift](https://docs.openshift.com/container-platform/3.11/architecture/networking/routes.html)
If you use different ports than `ROUTER_SERVICE_HTTP_PORT` and `ROUTER_SERVICE_HTTPS_PORT`, they are not taken into account

### DIFF
--- a/app/services/integration/kubernetes_service.rb
+++ b/app/services/integration/kubernetes_service.rb
@@ -135,18 +135,19 @@ class Integration::KubernetesService < Integration::ServiceBase
   class RouteSpec < K8s::Resource
     def initialize(url, service, port)
       uri = URI(url)
+      tls_options = {
+        insecureEdgeTerminationPolicy: 'Redirect',
+        termination: 'edge'
+      } if uri.class == URI::HTTPS
+
       super({
         host: uri.host || uri.path,
         port: { targetPort: port },
-        tls: {
-          insecureEdgeTerminationPolicy: 'Redirect',
-          termination: 'edge'
-        },
         to: {
           kind: 'Service',
           name: service
         }
-      })
+      }.merge(tls: tls_options))
     end
   end
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,8 +7,9 @@ jobs:
       - image: circleci/buildpack-deps:latest
         environment:
           POSGRES_CONTAINER_NAME: db
-          DATABASE_URL: postgresql://postgres:@db:5432/zync
+          DATABASE_URL: postgresql://postgres:postgres@db:5432/zync
           POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
           POSTGRES_DB: zync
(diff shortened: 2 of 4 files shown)


app/services/integration/kubernetes_service.rb
```diff
@@ -135,18 +135,19 @@ def initialize(attributes, **options)
...
-        tls: {
-          insecureEdgeTerminationPolicy: 'Redirect',
-          termination: 'edge'
-        },
         to: {
           kind: 'Service',
           name: service
         }
-      })
+      }.merge(tls: tls_options))
```
guicassolato: Is it OK to merge `tls: nil` if `uri.class != URI::HTTPS`?

hallelujah: Yes that is what k8s expect, there is a test for it

Anyway this PR is in conflict with the `KUBERNETES_ROUTE_TLS` env
I will think of a way to make it compatible later

hallelujah:
I need them to make it pass tests, and really opening 2 PR is not going to happen :D

PR #280, opened 2019-12-09
Allow disabling integrations

hallelujah:
Disable routes creations for Kubernetes Integration by setting:
DISABLE_K8S_ROUTES_CREATION=1

### DIFF
--- a/test/models/integration_test.rb
+++ b/test/models/integration_test.rb
@@ -13,4 +13,56 @@ class IntegrationTest < ActiveSupport::TestCase
     assert_equal [ keycloak ], Integration.tenant_or_model(nil, model)
     assert_equal [ one ], Integration.tenant_or_model(tenant, nil)
   end
+
+  test 'enabled is by integration' do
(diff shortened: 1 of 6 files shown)


guicassolato (approved):
I think we didn't have a test for `Integration#enabled?` before and now it seems even more important to have one IMHO. I'm OK with merging this if we need it soon though.

test/models/integration_test.rb
```diff
@@ -13,4 +13,56 @@ class IntegrationTest < ActiveSupport::TestCase
...
+
+  test 'enabled is by integration' do
+    with_integration keycloak: false, rest: true, kubernetes: true do
+      assert Integration.new.enabled?
+      refute Integration::Keycloak.new.enabled?
+    end
+  end
+
+  test 'rest enabled?' do
+      integration = Integration::REST.new
```
guicassolato, line 25:
```suggestion
    integration = Integration::REST.new
```

guicassolato (approved):
Perfect!

PR #314, opened 2020-03-30
Upgrade PostgreSQL client to 10

raelga:
Amazon Web Services no longer supports PostgreSQL 9.2, that reached End-of-Life in September, 2017. 

This PRs updates the PostgreSQL version by using the oficial PostgreSQL repository.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,30 +1,33 @@
 FROM registry.access.redhat.com/ubi7/ruby-25
 
 USER root
-RUN rpm -Uvh http://yum.postgresql.org/9.6/redhat/rhel-7-x86_64/pgdg-centos96-9.6-3.noarch.rpm \
+
   && yum update -y \
   && yum remove -y postgresql \
   && yum install -y postgresql96 postgresql96-devel postgresql96-libs \
   && yum clean all \
   && rm -rf /var/cache/yum
 
-RUN source ${APP_ROOT}/etc/scl_enable \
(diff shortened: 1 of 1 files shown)


hallelujah:
Lets use 10.6 as per support matrix in the last product version

image

raelga:
Bumped to 10 in https://github.com/3scale/zync/pull/314/commits/4ced0ffae81c9fe22adb53f195f6f65925abab64

raelga:
@hallelujah can you check this PR?

Dockerfile
```diff
@@ -1,20 +1,33 @@
...
-&& bundle exec bin/rails server -e production -d; \
-rm -rf tmp/pids
-USER root
-RUN chmod -fR g+w tmp/cache
-USER default
+  && bundle exec bin/rails server -e production -d; \
+  rm -rf tmp/pids
+
+RUN mkdir -p -m 0775 tmp/cache log \
+  && chown -fR default tmp log db \
```
hallelujah, line 30: Why do we need `db` here?

raelga: The pgdump commands fails otherwise with a permission denied on boot.

hallelujah: Maybe we should have the whole directory belonging to default user?

raelga: Just by giving permissions to those folders work as expected, It should be necessary to give extra permissions if they are not needed.

PR #315, opened 2020-04-02
Revert "Route termination options follow public url scheme"

guicassolato:
Reverts 3scale/zync#265 (except unrelated changes)

OpenShift insecure routes are incompatible with Rails's `force_ssl`, enforced by [3scale/porta](https://github.com/3scale/porta) at https://github.com/3scale/porta/blob/c67d5d8227a1f64c0ff2c0d91f5ea39a00cb9fb9/openshift/system/config/settings.yml#L5

Reverting the changes is just an emergencial measure. The proper way to fix this would be in [3scale/porta](https://github.com/3scale/porta) by making sure the route is communicated to zync as an https route.

### DIFF
--- a/app/services/integration/kubernetes_service.rb
+++ b/app/services/integration/kubernetes_service.rb
@@ -135,19 +135,18 @@ class Integration::KubernetesService < Integration::ServiceBase
   class RouteSpec < K8s::Resource
     def initialize(url, service, port)
       uri = URI(url)
-      tls_options = {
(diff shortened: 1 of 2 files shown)


hallelujah:
Can't we deploy the previous version of zync image instead? it would be quicker.

guicassolato:
Closed in favour if https://github.com/3scale/zync/pull/321

PR #318, opened 2020-05-05
[database.yml] schema_search_path public

macejmic:
Solution and problem explained in here: https://github.com/que-rb/que/issues/219

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,9 +1,5 @@
-# SQLite version 3.x
-#   gem install sqlite3
-#
-#   Ensure the SQLite 3 gem is defined in your Gemfile
-#   gem 'sqlite3'
-#
+schema_search_path: 'public'
+
 default: &default
   adapter: postgresql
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -1,9 +1,5 @@
-# SQLite version 3.x
-#   gem install sqlite3
-#
-#   Ensure the SQLite 3 gem is defined in your Gemfile
-#   gem 'sqlite3'
-#
+schema_search_path: 'public'
```
guicassolato, line 1: Not sure if we can enforce this. On Openshift, que uses `DATABASE_URL` to connect, which is defined in a secret, meaning the customer can eventually chose a different db schema.

I'd test first deploying to Openshift with zync's db in a different schema and see if this config is not in the way to establish the connection.

macejmic: it turned out this wasn't an issue at the end.. closing..

PR #328, opened 2020-05-14
Use UBI base image when building in Openshift

guicassolato:
This PR changes the ImageStream and BuildConfig YAML templates offered as examples to build zync from source on Openshift (s2i). Instead of using centos7 as base image, it uses UBI7 Ruby 2.5 image, i.e. same one of the Dockerfile.

### DIFF
--- a/openshift/02-ruby-25-ubi7-imagestream.yml
+++ b/openshift/02-ruby-25-ubi7-imagestream.yml
@@ -3,12 +3,12 @@ kind: ImageStream
 metadata:
   labels:
     app: zync
-  name: ruby-24-centos7
+  name: ruby-25-ubi7
 spec:
   tags:
   - from:
       kind: DockerImage
-      name: centos/ruby-24-centos7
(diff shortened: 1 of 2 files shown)


gsaslis:
Just to add that this is also going from ruby 2.4 to ruby 2.5. (i.e. it's not _just_ centos to ubi switch).

This is probably also fine as I remember prod images are also on ruby 2.5, just thought worth pointing out, in case anyone else thinks there might be sth wrong with that.

guicassolato:
> Just to add that this is also going from ruby 2.4 to ruby 2.5. (i.e. it's not _just_ centos to ubi switch).
> 
> This is probably also fine as I remember prod images are also on ruby 2.5, just thought worth pointing out, in case anyone else thinks there might be sth wrong with that.

That's what we have in the Dockerfile already, @gsaslis. CircleCI is also running based on Ruby 2.5. Thanks for pointing out anyway. It's important to highlight that.

PR #330, opened 2020-05-15
Switch CI image to UBI7-based ruby 2.5 image

gsaslis:
We should be running the tests in as close an environment as possible to the one used during productization


Martouta:
image

😞

gsaslis:
@Martouta i'll try to get this passing ;)

PR #331, opened 2020-05-15
Removes the flag chown from COPY commands for backward compatibility

guicassolato:
`--chown` was introduced in Docker v17.09 and won't work in previous versions, failing to build the image for example in older Openshift clusters.

This significantly increases the size of the image built from 719MB to 846MB. We could mitigate this effect a little by changing the order of steps in the Dockerfile therefore creating less layers. The downside of it is that we might be running useless heavy steps such as `COPY . .` (followed by `chown`) in cases where `RUN bundle install --deployment` fails for example.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,13 +14,20 @@ WORKDIR ${APP_ROOT}
 RUN source ${APP_ROOT}/etc/scl_enable \
   && gem install bundler --version=2.0.1 --no-document
 
-COPY --chown=default:root Gemfile* ./
+USER root
+COPY Gemfile* ./
+RUN chown -fR default:root ./Gemfile
+
+USER default
 RUN source ${APP_ROOT}/etc/scl_enable \
   && bundle config build.pg --with-pg-config=/usr/pgsql-10/bin/pg_config \
   && bundle install --deployment --path vendor/bundle --jobs $(grep -c processor /proc/cpuinfo) --retry 3
 
-COPY --chown=default:root . .
+USER root
+COPY . .
+RUN chown -fR default:root .
 
+USER default
 ENV RAILS_LOG_TO_STDOUT=1
 RUN source ${APP_ROOT}/etc/scl_enable \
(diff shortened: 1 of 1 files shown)


guicassolato:
@slopezz and I spoke and we decided maybe NOT to move on with this change. This was first motivated because of old running clusters that fail to build Zync from source with the current Dockerfile. However, those clusters should not be doing so anyway.

Whoever wants to build in Openshift based on older Kubernetes versions can always adapt the Dockerfile. Moreover, Red Hat offers proper downstream images for production usage.

We're keeping this on hold for a little longer and then likely closing it for good.

Dockerfile
```diff
@@ -14,13 +14,20 @@ WORKDIR ${APP_ROOT}
...
+
+USER default
 RUN source ${APP_ROOT}/etc/scl_enable \
   && bundle config build.pg --with-pg-config=/usr/pgsql-10/bin/pg_config \
   && bundle install --deployment --path vendor/bundle --jobs $(grep -c processor /proc/cpuinfo) --retry 3
 
-COPY --chown=default:root . .
+USER root
+COPY . .
+RUN chown -fR default:root .
```
hallelujah, line 29: It would be nice to group them to this section https://github.com/3scale/zync/pull/331/files#diff-3254677a7917c6c01f55212f86c57fbfR17

Then having less layer

guicassolato: > It would be nice to group them to this section https://github.com/3scale/zync/pull/331/files#diff-3254677a7917c6c01f55212f86c57fbfR17
> 
> Then having less layer

Yep. That's what I meant in the description with

> This significantly increases the size of the image built from 719MB to 846MB. We could mitigate this effect a little by changing the order of steps in the Dockerfile therefore creating less layers. The downside of it is that we might be running useless heavy steps such as `COPY . .` (followed by `chown`) in cases where `RUN bundle install --deployment` fails for example.

Nonetheless, we may end up closing this PR anyway.

PR #334, opened 2020-05-18
New Que stats Prometheus metrics

guicassolato:
- Renames the following measurement types of the `jobs_scheduled_total` Prometheus (gauge) metric:
  - `<job_class>/retry` to `<job_class>/retried` (Que jobs with at least 1 retry attempt)
  - `<job_class>/future` to `<job_class>/scheduled` (Que jobs scheduled to be executed in the future)

- Introduces the following new measurement types of the `jobs_scheduled_total` Prometheus (gauge) metric:
  - `<job_class>/ready`: Que jobs ready to be executed for the first time (excludes: errored, expired, finished, performed)
  - `<job_class>/finished`: Que jobs already performed
  - `<job_class>/failed`: Que jobs that failed at least once

Related to https://github.com/3scale/infrastructure/issues/905

### DIFF
--- a/lib/prometheus/que_stats.rb
+++ b/lib/prometheus/que_stats.rb
@@ -17,12 +17,21 @@ module Prometheus
       end
     end
 
+    # ApplicationJob did not have to be here, but it's just harder to test otherwise because of ApplicationJob#delete_duplicates
+    JOB_CLASSES = %w[ApplicationJob ProcessEntryJob ProcessIntegrationEntryJob UpdateJob].inspect.tr('"', "'").freeze
+
     def job_stats(*filters)
       filter = "WHERE #{filters.join(' AND ')}" if filters.presence
       sql = <<~SQL
-        SELECT args->0->>'job_class' AS job_class, COUNT(*) as count
-        FROM que_jobs #{filter}
-        GROUP BY args->0->>'job_class'
+        WITH
+          jobs AS (SELECT unnest(array#{JOB_CLASSES}) AS job_class),
+          stats AS (
+            SELECT args->0->>'job_class' AS job_class, COUNT(*) as count
+            FROM que_jobs #{filter}
+            GROUP BY args->0->>'job_class'
+          )
+        SELECT jobs.job_class, COALESCE(stats.count, 0) AS count
+        FROM jobs LEFT JOIN stats ON jobs.job_class = stats.job_class
       SQL
 
       execute do |connection|
--- a/lib/prometheus/que_stats.rb
+++ b/lib/prometheus/que_stats.rb
@@ -17,8 +17,8 @@ module Prometheus
       end
     end
 
-    def job_stats(filter = nil)
-      filter = "WHERE #{filter}" if filter
+    def job_stats(*filters)
+      filter = "WHERE #{filters.join(' AND ')}" if filters.presence
       sql = <<~SQL
         SELECT args->0->>'job_class' AS job_class, COUNT(*) as count
         FROM que_jobs #{filter}
@@ -28,6 +28,26 @@ module Prometheus
       execute do |connection|
         connection.select_all(sql)
       end
+    end
+
+    def job_stats_ready
+      job_stats('error_count = 0', 'expired_at IS NULL', 'finished_at IS NULL', 'run_at <= now()')
+    end
+
+    def job_stats_scheduled
+      job_stats('run_at > now()')
+    end
+
+    def job_stats_finished
+      job_stats('finished_at IS NOT NULL')
+    end
+
+    def job_stats_retried
+      job_stats(%q[(args->0->>'retries')::integer > 0])
+    end
+
+    def job_stats_failed
+      job_stats('error_count > 0')
     end
 
     mattr_accessor :read_only_transaction, default: true, instance_accessor: false
@@ -54,23 +74,17 @@ end
 Yabeda.configure do
   group :que do
     scheduled_jobs = gauge :jobs_scheduled_total, comment: 'Que Jobs to be executed'
-    workers = gauge :workers_total,  comment: 'Que Workers running'
+    workers = gauge :workers_total, comment: 'Que Workers running'
+
+    set_stats = ->(type, stats) { scheduled_jobs.set({ job: stats.fetch('job_class') }.merge(type), stats.fetch('count')) }
 
     collect do
-      Prometheus::QueStats.job_stats.each do |stats|
-        scheduled_jobs.set({ job: stats.fetch('job_class') }, stats.fetch('count'))
-      end
+      set_stats_all = set_stats.curry.call({})
+      Prometheus::QueStats.job_stats.each(&set_stats_all)
 
-      Prometheus::QueStats.job_stats(%q[(args->0->>'retries')::integer > 0]).each do |stats|
-        scheduled_jobs.set({ job: stats.fetch('job_class'), type: 'retry' }, stats.fetch('count'))
-      end
-
-      Prometheus::QueStats.job_stats('run_at < now()').each do |stats|
-        scheduled_jobs.set({ job: stats.fetch('job_class'), type: 'scheduled' }, stats.fetch('count'))
-      end
-
-      Prometheus::QueStats.job_stats('run_at > now()').each do |stats|
-        scheduled_jobs.set({ job: stats.fetch('job_class'), type: 'future' }, stats.fetch('count'))
+      %w[ready scheduled finished retried failed].each do |type|
+        set_stats_type = set_stats.curry.call(type: type)
+        Prometheus::QueStats.public_send("job_stats_#{type}").each(&set_stats_type)
       end
 
       workers.set({ }, Prometheus::QueStats.worker_stats.fetch('workers'))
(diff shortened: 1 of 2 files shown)


guicassolato:
On OpenShift:
```
$ oc exec -it $(oc get pods | grep zync-que | grep Running | awk '{ print $1 }') -- bash -c 'curl http://localhost:9394/metrics | grep que_jobs_scheduled_total'

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  7480  100  7480    0     0   130k      0 --:--:-- --:--:-- --:--:--  132k
# TYPE que_jobs_scheduled_total gauge
# HELP que_jobs_scheduled_total Que Jobs to be executed
que_jobs_scheduled_total{job="ProcessIntegrationEntryJob"} 1
que_jobs_scheduled_total{job="ProcessIntegrationEntryJob",type="scheduled"} 3
que_jobs_scheduled_total{job="ProcessIntegrationEntryJob",type="failed"} 3
que_jobs_scheduled_total{job="ProcessIntegrationEntryJob",type="ready"} 1
```

lib/prometheus/que_stats.rb
```diff
@@ -17,19 +17,48 @@ def worker_stats
...
+    JOB_CLASSES = %w[ApplicationJob ProcessEntryJob ProcessIntegrationEntryJob UpdateJob].inspect.tr('"', "'").freeze
+
+    def job_stats(*filters)
+      filter = "WHERE #{filters.join(' AND ')}" if filters.presence
       sql = <<~SQL
-        SELECT args->0->>'job_class' AS job_class, COUNT(*) as count
-        FROM que_jobs #{filter}
-        GROUP BY args->0->>'job_class'
+        WITH
+          jobs AS (SELECT unnest(array#{JOB_CLASSES}) AS job_class),
```
hallelujah, line 27: Can't we do a quote of an Array of String instead of doing the `inspect.tr` above?

lib/prometheus/que_stats.rb
```diff
@@ -17,19 +17,48 @@ def worker_stats
       end
     end
 
-    def job_stats(filter = nil)
-      filter = "WHERE #{filter}" if filter
+    # ApplicationJob did not have to be here, but it's just harder to test otherwise because of ApplicationJob#delete_duplicates
+    JOB_CLASSES = %w[ApplicationJob ProcessEntryJob ProcessIntegrationEntryJob UpdateJob].inspect.tr('"', "'").freeze
+
+    def job_stats(*filters)
+      filter = "WHERE #{filters.join(' AND ')}" if filters.presence
```
hallelujah, line 24: Does that work?

```ruby
filters.each_with_object(scope) {|filter, condition| condition.where(filter) }
```

lib/prometheus/que_stats.rb
```diff
@@ -17,19 +17,48 @@ def worker_stats
...
       sql = <<~SQL
-        SELECT args->0->>'job_class' AS job_class, COUNT(*) as count
-        FROM que_jobs #{filter}
-        GROUP BY args->0->>'job_class'
+        WITH
```
hallelujah, line 26: Common Table expression can be done in Arel, there is an open PR in https://github.com/rails/rails/pull/37944
I think you can do it without raw SQL

lib/prometheus/que_stats.rb
```diff
@@ -54,23 +83,17 @@ def execute
...
-      Prometheus::QueStats.job_stats(%q[(args->0->>'retries')::integer > 0]).each do |stats|
-        scheduled_jobs.set({ job: stats.fetch('job_class'), type: 'retry' }, stats.fetch('count'))
-      end
+    set_stats = ->(type, stats) { scheduled_jobs.set({ job: stats.fetch('job_class') }.merge(type), stats.fetch('count')) }
 
-      Prometheus::QueStats.job_stats('run_at < now()').each do |stats|
-        scheduled_jobs.set({ job: stats.fetch('job_class'), type: 'scheduled' }, stats.fetch('count'))
-      end
+    collect do
+      set_stats_all = set_stats.curry.call({})
```
hallelujah, line 91: Yes `type` is definitely a `Hash` ... this is unexpected. If the reader does not know how `curry` works, one can be confused

lib/prometheus/que_stats.rb
```diff
@@ -54,23 +83,17 @@ def execute
...
-    collect do
-      Prometheus::QueStats.job_stats.each do |stats|
-        scheduled_jobs.set({ job: stats.fetch('job_class') }, stats.fetch('count'))
-      end
+    workers = gauge :workers_total, comment: 'Que Workers running'
 
-      Prometheus::QueStats.job_stats(%q[(args->0->>'retries')::integer > 0]).each do |stats|
-        scheduled_jobs.set({ job: stats.fetch('job_class'), type: 'retry' }, stats.fetch('count'))
-      end
+    set_stats = ->(type, stats) { scheduled_jobs.set({ job: stats.fetch('job_class') }.merge(type), stats.fetch('count')) }
```
hallelujah, line 88: `type` is a `Hash`? Maybe rename it?

lib/prometheus/que_stats.rb
```diff
@@ -17,19 +17,48 @@ def worker_stats
...
+        FROM jobs LEFT JOIN stats ON jobs.job_class = stats.job_class
       SQL
 
       execute do |connection|
         connection.select_all(sql)
       end
     end
 
+    def job_stats_ready
+      job_stats('error_count = 0', 'expired_at IS NULL', 'finished_at IS NULL', 'run_at <= now()')
```
hallelujah, line 43: Could be:

```ruby
# You can keep last condition instead of using Time.zone.now if you want
conditions = [ {error_count: 0, expired_at: nil, finished_at: nil}, ['run_at <= ?', Time.zone.now]]

job_stats(conditions)

```

lib/prometheus/que_stats.rb
Martouta: I had to do something similar once but in BabySqueel and using OR instead of AND, but in case it helps:

https://github.com/3scale/porta/blob/ec3962a0b7de9357afa280bde39c90965a19e93b/app/models/cinstance.rb#L181-L188

lib/prometheus/que_stats.rb
Martouta: I am that confused reader unfortunately 🙈 Sorry 🙏 

And what is `curry` ? 😱

lib/prometheus/que_stats.rb
hallelujah: :+1: Yes and baby_squeel is installed

lib/prometheus/que_stats.rb
hallelujah: @Martouta It is a feature of functional programming. The best ever post I read about it was on SO https://stackoverflow.com/questions/36314/what-is-currying

Martouta: Oh, I see. Thanks! ❤️ I got to know about this functional programming feature thanks to Didier when he taught us a bit about Elm, but I didn't know that this feature had the name _curry_ 😄

lib/prometheus/que_stats.rb
guicassolato: IDK for sure why this was originally done in raw SQL. Maybe the intention was for it to depend as little as possible in the Rails environment and other dependencies of the application. Who knows?

However, if we want to change this, there's actually much more than we can do here. We can use Que's scopes: https://github.com/que-rb/que/blob/a2c9a2d7a11debee4ce5bd6121a4a5e0f6fd3323/lib/que/active_record/model.rb#L10-L23. Anyway, I'd prefer making this a "refactoring" PR following this one, so we can move on with configuring the dashboard and putting the new metrics in production ASAP.

lib/prometheus/que_stats.rb
guicassolato: Well, people not knowing how a core Ruby method works should not be an excuse for not using it, no? However, I did think about doing this completely different BTW, for stylistic reasons. We could have a proper class that fetches the count and sets the stats, with proper methods instead using procs and currying.

lib/prometheus/que_stats.rb
guicassolato: Following https://github.com/3scale/zync/pull/334/files#r429103152.

lib/prometheus/que_stats.rb
Martouta: @guicassolato The request is not to stop using it, the request is changing the name `type` to make more clear what it is 😄 

https://github.com/3scale/zync/pull/334/files#r429113867 & https://github.com/3scale/zync/pull/334/files#r429113547 are very related

lib/prometheus/que_stats.rb
Martouta: That's a good point 👍 It is probably safer to do it in 2 PRs to don't have too many risky changes in only 1.

lib/prometheus/que_stats.rb
guicassolato:
```
can't quote Array
```

PR #348, opened 2020-07-14
Fix race condition between entry jobs trying to create same proxy integration

guicassolato:
`ApplicationRecord.retry_record_not_unique` returns `nil` when it fails with `ActiveRecord::RecordNotUnique` more than once, which can happen in case of a race condition involving 3+ jobs trying to process entries that maps to the same tenant/integration type. Particularly, this makes https://github.com/3scale/zync/blob/d1d6360de9064adcf5e0e3c7b810b381edc4a87a/app/jobs/process_entry_job.rb#L27 to fail with:

```ruby
NoMethodError: undefined method `each' for nil:NilClass
```

----

Closes [THREESCALE-5632](https://issues.redhat.com/browse/THREESCALE-5632)

### DIFF
--- a/app/jobs/process_entry_job.rb
+++ b/app/jobs/process_entry_job.rb
@@ -13,18 +13,20 @@ class ProcessEntryJob < ApplicationJob
   end
 
   def self.model_integrations_for(entry)
+    self.ensure_integrations_for(entry)
+
     model = entry.model
-
-    integrations = Integration.retry_record_not_unique do
-      case model.record
-      when Proxy then self.const_get(:PROXY_INTEGRATIONS).map { |i| i.new(entry) }.each(&:call)
-      when Provider then CreateK8SIntegration.new(entry).call
-      end
-
-      Integration.for_model(model)
-    end
-
+    integrations = Integration.for_model(model)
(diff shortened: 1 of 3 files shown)


app/jobs/process_entry_job.rb
```diff
@@ -13,18 +13,20 @@ def perform(entry)
...
+    model = entry.model
+    integrations = Integration.for_model(model)
+    integrations.each.with_object(model)
+  end
 
-      Integration.for_model(model)
+  def self.ensure_integrations_for(entry)
+    case entry.model.record
+    when Proxy
+      self.const_get(:PROXY_INTEGRATIONS).map { |integration| integration.new(entry) }.each(&:call)
```
hallelujah, line 26: Why this `const_get` ?

guicassolato: I wanted to override it in the test
https://github.com/3scale/zync/blob/347ee9a2e9afc4dda6af95675564b8f9c5a22758/test/jobs/process_entry_job_test.rb#L70

mikz: Better to use class_attribute then?

PR #402, opened 2020-11-03
Merge pull request #375 from 3scale/gem/update-codecov

guicassolato:
[gem] Update codecov

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -81,10 +81,9 @@ GEM
     bugsnag-capistrano (1.1.2)
     builder (3.2.4)
     byebug (11.0.0)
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -369,4 +366,4 @@ DEPENDENCIES
   yabeda-rails
 
 BUNDLED WITH
-   2.0.2
```
eguzki, line 372: This is ok, but JFYI the bundler version of the productization process is 2.0.2. So it will complain. 

In the toolbox we also `Bundled With` 2.1.4 with no issues

eguzki (approved):

PR #444, opened 2021-04-27
Adding activerecord-pg_enum gem

HemaHG:
Adding ActiveRecord::PGEnum Gem to support `enum`

**Issue**
undefined method `create_enum' error while migrating in production

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -37,8 +37,6 @@ $$;
 
 SET default_tablespace = '';
 
-SET default_with_oids = false;
-
 --
 -- Name: que_jobs; Type: TABLE; Schema: public; Owner: -
 --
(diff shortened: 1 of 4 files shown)


hallelujah (dismissed):

hallelujah (approved):

db/structure.sql
```diff
@@ -1178,10 +1176,10 @@ CREATE INDEX index_metrics_on_tenant_id ON public.metrics USING btree (tenant_id
 
 
 --
--- Name: index_models_on_record_type_and_record_id; Type: INDEX; Schema: public; Owner: -
+-- Name: index_models_on_record; Type: INDEX; Schema: public; Owner: -
 --
 
-CREATE UNIQUE INDEX index_models_on_record_type_and_record_id ON public.models USING btree (record_type, record_id);
+CREATE UNIQUE INDEX index_models_on_record ON public.models USING btree (record_type, record_id);
```
hallelujah, line 1182: @HemaHG 

## Concern

Those SQL changes need to be documented  if no further work is done. My concern is that we would need to have migration path and documentation for existing customers to change this index name, and that is quite an unnecessary work for @3scale/documentation people and for customers while we can automatize it.


## The better solution is:

@HemaHG you can also make a migration that will change the name of this index. In the migration:

- delete the index `index_models_on_record_type_and_record_id` if it exists
- add the new index if it does not exist


The only documentation needed is for any upgrade the `zync-*` pods should be stopped. And it is already documented, so nothing to do in terms of doc. JFYI

hallelujah (approved):
You can make the migration for the index in another PR ;)

db/structure.sql
HemaHG: I will create a PR for migration

PR #447, opened 2021-05-04
removing index on remove_index_models_on_record_type_and_record_id and adding new index

HemaHG:
Addressing the review comments: https://github.com/3scale/zync/pull/444#discussion_r622898857


db/migrate/20210504152609_add_index_models_on_record.rb
```diff
@@ -0,0 +1,7 @@
+class AddIndexModelsOnRecord < ActiveRecord::Migration[6.1]
+  def change
+    create_table :models, if_not_exists: true do |t|
```
hallelujah, line 3: You do not want to recreate the table

hallelujah (changes requested):

db/migrate/20210504151301_remove_index_models_on_record_type_and_record_id.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveIndexModelsOnRecordTypeAndRecordId < ActiveRecord::Migration[6.1]
+  def change
+    remove_index :models, name: "index_models_on_record_type_and_record_id" , if_exists: true
+  end
+end
```
hallelujah, line 5: You can merge both migrations into one

db/migrate/20210504152609_add_index_models_on_record.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexModelsOnRecord < ActiveRecord::Migration[6.1]
+  def change
+    add_index :models, [:record_id, :record_type], name: "index_models_on_record", if_not_exists: true
```
hallelujah, line 3: The uniqueness constraints is gone :(

hallelujah (changes requested):

db/migrate/20210504151301_remove_index_models_on_record_type_and_record_id.rb
HemaHG: I thought the same but was not sure. So created two migrations. will merge both :)

hallelujah (approved):

PR #452, opened 2021-05-14
.env support and readme update

akostadinov:


hallelujah (changes requested):
I'd prefer not adding unecessary dependencies to the project:

- some people prefer configuration
- other prefer environment
- other prefer default settings (database, and such)
At last it is only a matter of taste, nothing to do with the project. You can add example files but not dependencies for such things I think. Especially `.env` file can be used by another tool that hooks into your shell than `dotenv-rails`

Personally I use [direnv](https://github.com/direnv/direnv) which is quite powerful.

akostadinov:
@hallelujah , would `direnv` work with scripts/tools like the [porta-dev-tools](https://github.com/guicassolato/porta-dev-tools) project? On a first glance it appears to be working only when you `cd` with your shell. Please correct me if I'm wrong.

I added `dotenv` gem because I saw it used in porta and I spoke with @guicassolato that it is fine to add it. Will it make any difference to you if I add `.env` only to test and dev deps?

Otherwise I can add `.env` processing to `ports-dev-tools` script. Just might be somehow less convenient for direct usage (without porta-dev-tools).

WDYT?

akostadinov:
Or add `.env` handling in `config/environment.rb` although this may not be very different from using the gem?

Gemfile
```diff
@@ -39,6 +39,8 @@ gem 'bugsnag'
 # https://github.com/bugsnag/bugsnag-capistrano/blob/8bcfb27cf6eaff312eef086cce729d553a431460/UPGRADING.md
 gem 'bugsnag-capistrano', '< 2', require: false
 
+gem 'dotenv-rails', '~> 2.7'
```
hallelujah, line 42: we already have it btw ... https://github.com/3scale/porta/blob/2a05a51a6ef35967f3c438ceff3a5d548f5bcd6a/Gemfile#L16

But I really want to remove it. I think it was used at the premise of the project ...

I do not believe an app should use a file to load the environments it needs ... This concept is already strange for the least...

I would understand if it is a launcher but not the app itself... 

For example unicorn zero-downtime restart used to not work well with those .env files
The author of dotenv really advises against using it in production anyway https://github.com/bkeepers/dotenv/issues/176#issuecomment-81792965
So at least use this only for development only. For test, I do not want to use it in a CI ... debugging can be hard.

So do you still want to use it in development?
I'd say you have other choices, like [direnv](https://github.com/direnv/direnv)  or [related projects](https://github.com/direnv/direnv#related-projects) 

Or just a (profile/zsh.bash.fish).rc like ... that you source before calling your executable ...
This is up to you to use, nothing coercing others and having hours of debugging session ending up with an env variable setup in a file not shown with `env` command ...

README.md
```diff
@@ -49,6 +49,24 @@ Given the locking on the **model** there will be some parallelization, but also
...
+
+It is easy with Docker or Podman
+```
+sudo dnf install postgresql pspg
+docker run -d -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_DB=zync --name postgres10-zync docker.io/circleci/postgres:10.5-alpine
+```
+
+### Start Zync
+```
+export DATABASE_URL=postgresql://postgres:@localhost:5432/zync # or use .env
```
hallelujah, line 64: I'd rather avoid `.env`

I personally use envrc with direnv but I do not want people to use that unless they know what they are doing

hallelujah (changes requested):

hallelujah:
yes direnv is working when you "enter" the directory

But I would not either mention that in the project as something official. Environment management is up to one self, nothing to do with the project.

You can even create a simple shell wrapper or an alias to do what you need ... just that everybody has its own way.
And the concept of dotenv is really non-sense to me:

Basically `boot the app first to load the environments needed to boot the app ...`

README.md
```diff
@@ -49,6 +49,25 @@ Given the locking on the **model** there will be some parallelization, but also
 
 Because **Zync** will keep a **log** of **events** it will be able to replay changes and recover the last state just taking last revisions of each **model** and even remove the ones that have been created before but have been deleted.
 
+## Running locally
+
+### Run a PostgreSQL server
+
+It is easy with Docker or Podman
+```
+sudo dnf install postgresql pspg
```
hallelujah, line 58: The `pspg` part is optional, a matter of taste. Not something you must have for the project. It is a cosmetic nice addon pager. (I do not use it...) Still it does not belong to the project.

By the way postgresql client is not needed, but `libpq-devel`
```suggestion
sudo dnf install postgresql libpq-devel
```

I leave postgresql as it is a must-have client.

README.md
```diff
@@ -49,6 +49,25 @@ Given the locking on the **model** there will be some parallelization, but also
...
+docker run -d -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_DB=zync --name postgres10-zync docker.io/circleci/postgres:10.5-alpine
```
hallelujah, line 59: OK this is adding docker as a dependency for the project... Install docker section?

But to make it the OS way you can use modules:

```suggestion
sudo dnf module enable postgresql:10
sudo dnf module switch-to postgresql:10
```

hallelujah (changes requested):

README.md
akostadinov: In porte we have MySQL instructions for container as well local installation. I prefer and have tested container installation of postfres, that's why I add only it. If you have tested OS installation, I can add it as well. Let me know.

README.md
akostadinov: Thanks, added. I guess I had it already because I didn't notice it missing.

README.md
hallelujah: Container is always an option but as you are using Fedora 34 in this section I'd continue with pure OS installation. Just to be consistent.

hallelujah: By the way in Fedora it is podman not docker thus the section about installation of Docker :D 
I guess it is just preferable to use the system OS

akostadinov: I'm not trying to write complete documentation but just add something that I am using and see working. 

btw documentation is not Fedora or Linux specific at all. It would need more work to make sections in readme and make it OS specific. For me `docker` example is a good improvement and should work across different OSes (including Fedora and `docker` alias for podman).

Adding Fedora and other OS specific information can be a future task IMO. Can we merge as is and leave further improvements for the future?

hallelujah: Well there is https://github.com/3scale/zync/blob/master/INSTALL.md

You should put it there by the way instead of in the README

And in this file we already have the installation procedure for Mac OS X with a Brewfile
So I assumed this is for Fedora

hallelujah: Another approach is just assume the user knows how to install different dependencies for its OS and knows how to run rails we can just list them:

- postgresql
- ruby 2.7

And it should be enough

akostadinov:
Please check again when you can. Thanks!

INSTALL.md
```diff
@@ -1,17 +1,55 @@
...
+```shell
+sudo dnf module install postgresql:10
+sudo dnf install libpq-devel
+sudo /usr/bin/postgresql-setup --initdb
+systemctl enable postgresql
+systemctl start postgresql
+psql -U postgres -c "create user ${USER}"
+psql -U postgres -c "create database zync_development with owner ${USER}"
+psql -U postgres -c "create database zync_test with owner ${USER}"
+psql -U postgres -c "create database zync_production with owner ${USER}"
```
hallelujah, line 29: You do not need this... Postgresql creation of user and db is known/documented by them and standard
https://www.postgresql.org/docs/current/app-createdb.html

I really do not think we should document other 3rd party software ... it looks like Postgresql guide more than zync ... and maintaining it will be tedious. 

I would prefer you just put a sentence:

> Setup a postgresql role with `createdb` and `login` grants

I really prefer assuming people know at least the basic of postgresql than documenting all for them because otherwise we will have to teach them rails, sql etc ... 

Hint how to do it on Fedora (no need to document here): 

```
sudo su - postgres
createuser <your_username> --createdb --login
exit
```

Then `./bin/setup` will create all the things for you

INSTALL.md
```diff
@@ -1,17 +1,55 @@
...
-4. Start postgres.
-`brew services start postgresql`
+## Setup PostgreSQL as a container with Docker or Podman
 
-5. Setup Zync.
+```
+docker run -d -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_DB=zync --name postgres10-zync docker.io/circleci/postgres:10.5-alpine
+```
+
+When running any Zync command, make sure to have `DATABASE_URL` environment variable set in prior.
```
hallelujah, line 38: Why so? the `config/database.yml` is enough and it is Rails convention, no need to tell people to use that particularly over the config file

INSTALL.md
```diff
@@ -1,17 +1,55 @@
...
+
+```
+export DATABASE_URL=postgresql://postgres:@localhost:5432/zync
+```
+
+**Note:** You will also have to install on the host machine `psql` client tool (needed for `db:setup`) and `libpq-devel` (needed to build `pg` gem).
```
hallelujah, line 44: I won't put that in a note, this is really valuable information and need proper instruction to setup the application.

INSTALL.md
```diff
@@ -1,17 +1,55 @@
...
+## Setup Zync
 `./bin/setup`
 
-The command `brew services start postgresql` starts the service of PostgreSQL. If `./bin/setup` aborts, make sure that the `PostgreSQL` service is running. Verify with `brew services list` that has a status `started` and looking green. If the status is `started` but coloured orange, fix the errors indicated in the log located in `/usr/local/var/log/postgres.log`
+
+## Start Zync
+```
+export ZYNC_AUTHENTICATION_TOKEN=token # must match porta config
+bundle exec rails server -p 5000
+PROMETHEUS_EXPORTER_PORT=9395 bundle exec rake que
```
hallelujah, line 54: About the environment variables I would prefer a documentation like for Apisonator or APIcast

https://github.com/3scale/apisonator#configuration
https://github.com/3scale/APIcast/blob/master/doc/parameters.md

INSTALL.md
```diff
@@ -1,17 +1,55 @@
...
+
+There is `Brewfile` containing all the dependencies.
+
+```
+cd zync
+brew bundle
+brew services start postgresql
+```
+
+The command `brew services start postgresql` starts the service of PostgreSQL. If later `./bin/setup` aborts, make sure that the `PostgreSQL` service is running. Verify with `brew services list` that has a status `started` and looking green. If the status is `started` but coloured orange, fix the errors indicated in the log located in `/usr/local/var/log/postgres.log`
```
hallelujah, line 16: This is more documenting `brew` than zync project

akostadinov: This comes from original text. Shall I remove it?

INSTALL.md
akostadinov: My intention was to write a few commands that worked for me to setup the environment without having to figure it out each time. If you think there is no place for such instructions in the repo, I can remove and keep locally or put in porta-dev-tools or wherever.

INSTALL.md
akostadinov: This is same as above - trying to provide ready commands to copy.paste. As a quick start guide. If you don't like it, I can remove and keep elsewhere.

INSTALL.md
akostadinov: I don't have executable `createdb` on my Fedora. I just put here the commands that worked for me. Same about copy/paste instructions like on the other two places applies here.

INSTALL.md
akostadinov: will update once direction of the pr is more clear

INSTALL.md
hallelujah: You need to login with `postgres` user to have it ;)

INSTALL.md
hallelujah: I think this is a common knowledge with Rails application. My remark is about the enforcing the user to use `DATABASE_URL` I do not think this should be the way because there is anotherway, editing`config/database.yml`

How about you write a wiki page "Quick start guide on Fedora 34" ? Or another Markdown file.

I would avoid to describe every cases in the README, it might get complicated to maintain it then.
My point is differentiating Zync purpose and other things.

INSTALL.md
hallelujah: Do not get me wrong, the quickstart guide is awesome for me but propably out of the scope of the README file.
Suggestion: put it in another file or better in a wiki page :)

INSTALL.md
hallelujah: Yes I think you should remove it. Lets try to focus on this README to make it so that people know about the project.
Then we can probably add some other `doc/*.md` files or just wiki page for some use cases.

akostadinov:
Please check again if my refactoring makes sense to you.

hallelujah (approved):

hallelujah (approved):

PR #456, opened 2021-06-04
Build latest image and push to quay.io as GH action

hallelujah:
Zync should build its own image and push to quay.io/3scale/zync:latest

:warning: DEPRECATION `nightly` tag is deprecated
and will be removed without further notice after 3 months

@3scale/support @3scale/documentation @3scale/qe If any documentation mentions nightly build please correct it

We are for now going to keep nightly tag for backward compatibility but it will reference the latest build

### DIFF
--- a/.github/workflows/quay-io-build-latest.yml
+++ b/.github/workflows/quay-io-build-latest.yml
@@ -1,0 +1,31 @@
+name: Build and push to quay.io the latest code
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


gsaslis:
Interesting... how come we're adding GitHub actions ? doesn't it make more sense to keep all CI on one system?

hallelujah:
@gsaslis The choice is just by me... But I had some reasons:

- I wanted to use our credits of GH actions
- I do not want to add a 15' job for each merge in CircleCI. Especially to reduce the cost and also to free some parallel job there.  Those can be used on other PR work
- It was straight-forward to set it up.

hallelujah:
Closing, the build will be done automatically in quay.io

PR #457, opened 2021-06-18
Support Postgresql 10 and 12

hallelujah:
Transitioning to 12 as the stable version

The community image is now built on top of UBI8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,15 @@
+.PHONY: build push help
+.DEFAULT_GOAL := help
+
+IMAGE_NAME := quay.io/3scale/zync
+TAG := latest
+IMAGE_TAG := $(IMAGE_NAME):$(TAG)
+
+build:
+	docker build . --tag $(IMAGE_TAG)
+
+push:
+	docker push $(IMAGE_TAG)
+
+help: ## Print this help
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   docker-build:
     resource_class: small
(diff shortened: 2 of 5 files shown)


Rakefile
```diff
@@ -6,3 +6,25 @@ require 'bugsnag-capistrano/tasks'
 require_relative 'config/application'
 
 Rails.application.load_tasks
+
+namespace :db do
```
guicassolato, line 10: Maybe in https://github.com/3scale/zync/blob/master/lib/tasks/db.rake?

Rakefile
```diff
@@ -6,3 +6,25 @@ require 'bugsnag-capistrano/tasks'
...
+    ActiveRecord::Tasks::PostgreSQLDatabaseTasks.new(db_config).set_schema_file
+  end
+end
+
+ActiveRecord::Tasks::PostgreSQLDatabaseTasks.prepend(Module.new do
+  def set_schema_file
+    clear_active_connections!
+    establish_master_connection
+    server_version = connection.select_value("SHOW server_version").to_i
+    suffix = server_version < 11 ? '' : '-12'
```
guicassolato, line 22: If it's psql 11, we still append `-12`? I guess the semantics then is "schema is meant for psql v12", even if the customer uses with psql 11.

Makefile
```diff
@@ -0,0 +1,15 @@
+.PHONY: build push help
+.DEFAULT_GOAL := help
+
+IMAGE_NAME := quay.io/3scale/zync
```
guicassolato, line 4:
```suggestion
IMAGE_NAME ?= quay.io/3scale/zync
```

Rakefile
hallelujah: Well because of the version used in Dockerfile and other scripts ... I do not know I would like just to migrate to 12 and not having backward compatibility

Makefile
guicassolato: Or maybe even in just one step:

```suggestion
ZYNC_IMAGE := quay.io/3scale/zync:latest
```

hallelujah: Why? I do not want to deffer the expansion if any here

guicassolato: To be able to control the image name from the outside. To open it up for other usage, like in dev workflows as well. E.g.:

```sh
ZYNC_IMAGE=quay.io/user/zync:my-branch make build
```

(Just a suggestion.)

hallelujah: hmmm ... what is the difference, it is a Makefile, you pass the args in the command line in your script:

```shell
make -f zync/Makefile IMAGE_NAME=${ZYNC_IMAGE}
```

guicassolato: You're totally right. Apologies.

guicassolato: I sometimes leverage exporting variables to the shell once and then running a series of make targets without carrying about the parameters any more. But you don't need that in this case, so never mind.

raelga (approved):
Deployed in stg-saas, triggered a migration (from 10 to 12.5 in https://github.com/3scale/platform/commit/3554701cfb2f360bc3b0322a5c22802072dcf8b5) and validated by @hallelujah.

/lgtm

PR #459, opened 2021-07-13
Fix Kubernetes service finding from owner references without metadata

guicassolato:
Kubernetes `meta/v1.OwnerReference` resources do not contain a `metadata` field and therefore Zync's `Integration::KubernetesService#owner_reference_controller` should not expect the `OpenStruct` Ruby objects that represent those resources to respond to `#metadata` and then recursively to `#[]=(value)`. Doing such will result in 

```ruby
NoMethodError: undefined method `[]=' for nil:NilClass
```

However, https://github.com/kontena/k8s-client still expects any resource reference passed as argument to `K8s::Client#get_resource` to respond to `#metadata`. To set the `metadata` field on the owner reference resource, we need to use `K8s::Resource#merge`

Closes [THREESCALE-7283](https://issues.redhat.com/browse/THREESCALE-7283)

### DIFF
--- a/app/services/integration/kubernetes_service.rb
+++ b/app/services/integration/kubernetes_service.rb
@@ -53,9 +53,8 @@ class Integration::KubernetesService < Integration::ServiceBase
   def owner_reference_controller(resource)
     owner_references = resource.metadata.ownerReferences or return
     controller = owner_references.find(&:controller)
-    controller.metadata = { namespace: namespace, name: controller.name }
 
-    client.get_resource(controller)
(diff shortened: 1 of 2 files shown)


makikata:
Correct see https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference

> OwnerReference contains enough information to let you identify an owning object. An owning object must be in the same > namespace as the dependent, or be cluster-scoped, so there is no namespace field.

makikata (approved):

hallelujah (approved):

PR #460, opened 2021-07-28
Fix custom ActiveJob::Base.rescue_from after upgrading to Rails 6

guicassolato:
Zync `UpdateJob` retries on `Errno::ECONNREFUSED` and `Model::LockTimeoutError` exceptions have been broken since we upgraded to Rails 6 (#440).

This is the case, e.g., when `UpdateState` fails to acquire a lock for a model and raises a `Model::LockErrorTiemout` at https://github.com/3scale/zync/blob/38d145ead0053799eb5f2c4fdb169ac57394d1b6/app/models/model.rb#L38

Our custom implementation for `ActiveJob::Base.retry_on` (to add debug log level) did not comply with the new Rails 6 API for the `ActiveJob::Base.retry_job` method, causing an `ArgumentError` raised when failed jobs were attempted to be enqueued for retry, e.g.:

```
E, [2021-07-27T07:08:32.919387 #1] ERROR -- : {"lib":"que","hostname":"zync-que-1-fkmkx","pid":1,"thread":18925860,"event":"job_errored","job_id":3,"elapsed":0.385098051,"error":"ArgumentError: wrong number of arguments (given 1, expected 0; required keywords: seconds_or_duration_or_algorithm, executions)"}
```

This PR fixes the patch so it complies with the Rails API.

After the change:

Screenshot 2021-07-29 at 12 30 28

----

Closes [THREESCALE-7369](https://issues.redhat.com/browse/THREESCALE-7369)

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -5,15 +5,18 @@ require 'que/active_record/model'
 # Base class for all Jobs
 class ApplicationJob < ActiveJob::Base
   # Copied from ActiveJob::Exceptions, but uses debug log level.
-  def self.retry_on(exception, wait: 3.seconds, attempts: 5, queue: nil, priority: nil)
+  def self.retry_on(exception, wait: 3.seconds, attempts: 5, queue: nil, priority: nil, jitter: ActiveJob::Exceptions.const_get(:JITTER_DEFAULT))
     rescue_from exception do |error|
       if executions < attempts
         logger.debug "Retrying #{self.class} in #{wait} seconds, due to a #{exception}. The original exception was #{error.cause.inspect}."
-        retry_job wait: determine_delay(wait), queue: queue, priority: priority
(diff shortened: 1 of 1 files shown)


app/jobs/application_job.rb
```diff
@@ -5,15 +5,18 @@
 # Base class for all Jobs
 class ApplicationJob < ActiveJob::Base
   # Copied from ActiveJob::Exceptions, but uses debug log level.
-  def self.retry_on(exception, wait: 3.seconds, attempts: 5, queue: nil, priority: nil)
+  def self.retry_on(exception, wait: 3.seconds, attempts: 5, queue: nil, priority: nil, jitter: ActiveJob::Exceptions.const_get(:JITTER_DEFAULT))
```
guicassolato, line 8: Jitter was [introduced](https://github.com/rails/rails/pull/31872) as an option here in Rails 6.1. Adding it to the monkey patch as well only so we stick with the same API.

hallelujah (approved):
🚀

PR #467, opened 2021-11-02
[THREESCALE-7766] remove abandoned message bus

akostadinov:
fixes https://issues.redhat.com/browse/THREESCALE-7766

in the spirit of 3scale/porta#2687

I have not tested this as part of full installation yet.


db/migrate/20211102180650_remove_message_bus.rb
```diff
@@ -0,0 +1,30 @@
+class RemoveMessageBus < ActiveRecord::Migration[5.2]
+  def up
+    drop_table :message_bus
+  end
+
+  def down
+    ActiveRecord::Base.connection.execute """
```
akostadinov, line 7: I'm not sure we need this `#down` method at all. It can't restore any deleted data anyway. Also we can't use `MessageBus` methods to recreate the table because that class is not present anymore. So I just took a dump of the DDL. Should I remove it?

db/migrate/20211102180650_remove_message_bus.rb
```diff
@@ -0,0 +1,16 @@
...
+  def down
+    create_table "message_bus", force: :cascade do |t|
+      t.text "channel", null: false
+      t.text "value", null: false
+      t.datetime "added_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
+      t.index ["added_at"], name: "table_added_at_index"
+      t.index ["channel", "id"], name: "table_channel_id_index"
+      t.check_constraint "octet_length(value) >= 2", name: "message_bus_value_check"
+    end
+  end
```
akostadinov, line 15: I'm not sure we need this `#down` method at all. It can't restore any deleted data anyway. Also we can't use `MessageBus` methods to recreate the table because that class is not present anymore. So I just took a dump of the DDL.

hallelujah: Well this is called irreversible migration :) You can look in the code there are some of them ;)
It is totally fine to delete them

akostadinov:
Running locally, I see after creating a new provider:
```
[ActiveJob] [ProcessEntryJob] [f21a2113-9aa2-4077-8103-7d4062fda2b7] Performed P
rocessEntryJob (Job ID: f21a2113-9aa2-4077-8103-7d4062fda2b7) from Que(default) 
in 3.58ms
Processing by TenantsController#update as JSON
  Parameters: {"id"=>4, "endpoint"=>"http://new-provider-admin.3scale.localhost:
3000/", "access_token"=>"[FILTERED]", "tenant"=>{"id"=>4, "endpoint"=>"http://ne
w-provider-admin.3scale.localhost:3000/", "access_token"=>"[FILTERED]"}}
Started PUT "/notification" for 127.0.0.1 at 2021-11-03 21:47:35 +0200
  TRANSACTION (0.2ms)  BEGIN
  ↳ app/models/tenant.rb:9:in `block in upsert'
  Tenant Load (0.3ms)  SELECT "tenants".* FROM "tenants" WHERE "tenants"."id" = 
$1 LIMIT $2  [["id", 4], ["LIMIT", 1]]
  ↳ app/models/tenant.rb:9:in `block in upsert'
  TRANSACTION (0.3ms)  COMMIT
  ↳ app/models/application_record.rb:9:in `retry_record_not_unique'
Completed 201 Created in 5ms (Views: 0.2ms | ActiveRecord: 0.8ms | Allocations: 
2112)
```

I'm trying to run inside OpenShift with the update code. Or is above enough to validate it is processing updates?

db/migrate/20211102180650_remove_message_bus.rb
hallelujah: So what about it? like https://github.com/3scale/porta/blob/3017562449d3275840cae6afe338a01c1dcca899/db/migrate/20100809103633_remove_wrong_cinstance_of_softonic.rb#L8

hallelujah:
Image building you can do it via quay.io directly no? Or build locally?
Create a new build trigger temporarily and run it

But it is not really needed unless you touched something that tamper the boot process of the container for example, if it works locally I do not see it not working on a container.

hallelujah (changes requested):
Please make sure we are not going back to it.
This migration is irreversible

akostadinov:
As we discussed, we will keep message bus for easier debugging and monitoring of what tasks is Zync performing.

PR #490, opened 2023-01-19
fix activerecord security alert

akostadinov:
https://github.com/3scale/zync/security/dependabot/42


Gemfile
```diff
@@ -63,10 +63,7 @@ group :development, :test do
   gem 'pry-rescue'
   gem 'pry-stack_explorer'
 
-  gem 'license_finder', '~> 6.10'
-  gem 'license_finder_xml_reporter', git: 'https://github.com/3scale/license_finder_xml_reporter.git', tag: '1.0.0'
-  # rubyzip is a transitive depencency from license_finder with vulnerability on < 1.3.0
-  gem 'rubyzip', '>= 1.3.0'
+  gem 'license_finder', '~> 7.0.1'
```
jlledom, line 66: Why updating license finder?

akostadinov: Old one failed to work. And this is same as in porta. So I thought we can get rid of old stuff.

akostadinov: And I couldn't update the licenses report what we have to do when gems are changed. Just we need CI to check that.

PR #498, opened 2023-03-24
Upgrade que gem to v1.1.0 to fix deadlock issues

mayorova:
There is an epic "Zync doesn't create routes": https://issues.redhat.com/browse/THREESCALE-7657

While this may be caused by different reasons, it seems that in some cases restarting `zync-que` pod resolves the issue (e.g. https://issues.redhat.com/browse/THREESCALE-6835)

There are some common reports on the [que-rb](https://github.com/que-rb) repo reporting similar issues - workers just stop processing the job queue, specifically https://github.com/que-rb/que/issues/241 explains that there is some kind of a deadlock that causes this.

This (and other issues that may potentially cause deadlocks) is fixed in the version [1.0.0.beta5 (2021-12-23)|https://github.com/que-rb/que/blob/master/CHANGELOG.md#100beta5-2021-12-23], or in fact 1.0.0. The currently used version is `1.0.0.beta4`.

This PR upgrades `que` to `1.1.0` - it is the last version that does does not have any deprecations that would appear with the currently used Rails version. In order to upgrade to >= 1.2.0 and upgrade to Rails >= 7.0.4 is also recommended.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,10 @@ RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/
 USER default
 WORKDIR ${APP_ROOT}
 
-RUN gem install bundler --version=2.3.25 --no-document
-
 COPY --chown=default:root Gemfile* ./
+
+RUN BUNDLER_VERSION=$(awk '/BUNDLED WITH/ { getline; print $1 }' Gemfile.lock) \
+    && gem install bundler --version=$BUNDLER_VERSION --no-document
 
 RUN bundle config build.pg --with-pg-config=/usr/pgsql-12/bin/pg_config \
   && bundle install --deployment --path vendor/bundle --jobs $(grep -c processor /proc/cpuinfo) --retry 3
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN dnf --setopt=skip_missing_names_on_install=False,tsflags=nodocs --save \
 USER default
 WORKDIR ${APP_ROOT}
 
-RUN gem install bundler --version=2.2.19 --no-document
+RUN gem install bundler --version=2.3.25 --no-document
 
 COPY --chown=default:root Gemfile* ./
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,10 @@
 FROM registry.access.redhat.com/ubi8/ruby-27
 
 USER root
-RUN dnf --setopt=skip_missing_names_on_install=False,tsflags=nodocs --save \
-  && rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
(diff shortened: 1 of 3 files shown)


akostadinov (approved):
If there are no database migration steps needed for the upgrade, then this looks like a very nice upgrade. Thank you!

Dockerfile
```diff
@@ -1,18 +1,17 @@
...
+  && dnf remove -y postgresql* \
+  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y shared-mime-info postgresql12-12.13 postgresql12-devel-12.13 postgresql12-libs-12.13 \
   && dnf clean all \
   && rm -rf /var/cache/yum
 
 USER default
 WORKDIR ${APP_ROOT}
 
-RUN gem install bundler --version=2.2.19 --no-document
+RUN gem install bundler --version=2.3.25 --no-document
```
mayorova, line 14: That's the version specified in Gemfile.lock

Dockerfile
```diff
@@ -1,18 +1,17 @@
...
 
 USER root
-RUN dnf --setopt=skip_missing_names_on_install=False,tsflags=nodocs --save \
-  && rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
+RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
   && dnf update -y \
-  && dnf remove -y postgresql \
-  && dnf install -y shared-mime-info postgresql12 postgresql12-devel postgresql12-libs \
+  && dnf remove -y postgresql* \
+  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y shared-mime-info postgresql12-12.13 postgresql12-devel-12.13 postgresql12-libs-12.13 \
```
mayorova, line 7: Version 12.14 depends on `perl(IPC::Run)` which is not in the standard repos.
Need to install epel for this.

We need to tackle this at some point anyway, because we need to upgrade to Postgres 13 (see [THREESCALE-8545](https://issues.redhat.com/browse/THREESCALE-8545)), and postgres13 depends on it too.

jlledom (dismissed):

Dockerfile
akostadinov:
```suggestion
RUN BUNDLER_VERSION=$(awk '/BUNDLED WITH/ { getline; print $1 }' Gemfile.lock) \
    gem install bundler --version=$BUNDLER_VERSION --no-document
```

Dockerfile
akostadinov: we would maybe need to use CentOS Stream 8 in such case (in the future_)

Dockerfile
josemigallas: Which Gemfile.lock?
https://github.com/3scale/porta/blob/30f3232eee6da0d76a18076261be6bf4e4bb8622/Gemfile.lock#L1068-L1069

akostadinov (approved):
Changes are nice, thank you! We better use postgres from RH though in the future.

PR #503, opened 2023-06-26
Fix CircleCI builds

mayorova:
This PR does a few things:

- updates CircleCI base image, see https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034
- upgrades PostgreSQL image to 13. I don't know why there was 12 in the first place. [Supported Configurations](https://access.redhat.com/articles/2798521) doesn't specify what version of **external** PostgreSQL is supported for Zync, but it only mentions 13. The built-in database for Zync in 3scale operator is version 10, there is a JIRA to upgrade it:
[THREESCALE-8545: change support of PostgreSQL (internal db) from 10.x to 13.x](https://issues.redhat.com/browse/THREESCALE-8545)
- tweaks the Dockerfile to fix the failing build (see inline comments)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     resource_class: small
     working_directory: /home/circleci/zync
     docker:
-      - image: circleci/buildpack-deps:latest
+      - image: cimg/base:current
         environment:
           POSGRES_CONTAINER_NAME: db
           DATABASE_URL: postgresql://postgres:postgres@db:5432/zync
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
     working_directory: /opt/app-root/zync
     docker:
       - image: registry.access.redhat.com/ubi7/ruby-27
-      - image: circleci/postgres:<< parameters.postgresql_version >>-ram
+      - image: cimg/postgres:<< parameters.postgresql_version >>
     environment:
         RAILS_ENV: test
         DISABLE_SPRING: 1 # we can't really run spring as it hangs on local circleci build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       - setup_remote_docker
       - run: docker build --tag zync:build --file ./Dockerfile .
       - run: docker network create net0
-      - run: docker run --net net0 --name ${POSGRES_CONTAINER_NAME} -d -p 5432:5432 -e POSTGRES_USER=${POSTGRES_USER} -e POSTGRES_PASSWORD=${POSTGRES_PASSWORD} -e POSTGRES_DB=${POSTGRES_DB} postgres:12-alpine
+      - run: docker run --net net0 --name ${POSGRES_CONTAINER_NAME} -d -p 5432:5432 -e POSTGRES_USER=${POSTGRES_USER} -e POSTGRES_PASSWORD=${POSTGRES_PASSWORD} -e POSTGRES_DB=${POSTGRES_DB} postgres:13-alpine
       - run:
           command: |
             docker run --net net0 -e RAILS_ENV=${RAILS_ENV} -e DATABASE_URL=${DATABASE_URL} \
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -4,9 +4,9 @@ jobs:
     resource_class: small
     working_directory: /home/circleci/zync
     docker:
-      - image: circleci/buildpack-deps:latest
+      - image: cimg/base:current
         environment:
-          POSGRES_CONTAINER_NAME: db
+          POSTGRES_CONTAINER_NAME: db
```
mayorova, line 9: Just fixing a typo

Dockerfile
```diff
@@ -2,9 +2,7 @@ FROM registry.access.redhat.com/ubi8/ruby-27
 
 USER root
 RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
-  && dnf update -y \
-  && dnf remove -y postgresql* \
-  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y shared-mime-info postgresql12-12.13 postgresql12-devel-12.13 postgresql12-libs-12.13 \
+  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs --skip-broken -y shared-mime-info postgresql13 postgresql13-devel postgresql13-libs \
```
mayorova, line 5: `dnf update` was failing all the time with:

```
Error: 
 Problem: cannot install the best update candidate for package libpq-devel-13.5-1.el8.x86_64
  - nothing provides perl(IPC::Run) needed by postgresql15-devel-15.3-2PGDG.rhel8.x86_64
  - nothing provides perl(IPC::Run) needed by postgresql14-devel-14.8-2PGDG.rhel8.x86_64
  - nothing provides perl(IPC::Run) needed by postgresql13-devel-13.11-2PGDG.rhel8.x86_64
  - nothing provides perl(IPC::Run) needed by postgresql12-devel-12.15-2PGDG.rhel8.x86_64
  - nothing provides perl(IPC::Run) needed by postgresql11-devel-11.20-2PGDG.rhel8.x86_64
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
Error: error building at STEP "RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm   && dnf update -y   && dnf remove -y postgresql*   && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y shared-mime-info postgresql13 postgresql13-devel postgresql13-libs   && dnf clean all   && rm -rf /var/cache/yum": error while running runtime: exit status 1
```

I did not find a way to add the package providing `perl(IPC::Run)`

Also, installing postgresql13 was also failing with the same error without `--skip-broken`.
This doesn't seem to affect the tests. We probably need to check whether the image built by this Dockerfile actually works well. But if the `rails db:setup` step succeeds I don't know why it wouldn't.

josemigallas (approved):

.circleci/config.yml
```diff
@@ -90,5 +90,5 @@ workflows:
       - build:
           matrix:
             parameters:
-              postgresql_version: [ "10-alpine", "12-alpine" ]
+              postgresql_image: [ "circleci/postgres:10-alpine-ram", "cimg/postgres:13.11" ]
```
akostadinov, line 93: Why do we use `postgres:13-alpine` above and here we use the circleci images? Just in case there is a special reason. I guess because we used to do in this way.

Dockerfile
akostadinov: I think that we only need postgres libraries here, moreover the postgres libraries that come with RHEL, not external ones that are not used in production.

.circleci/config.yml
mayorova: Yeah, that surprised me also, I don't know why! It was already like that... 

Happy to change `postgres:13-alpine` to a CircleCI image, if it makes sense.

We'll need to add version 12 in the tests, because I became aware that this is the version we still use in SaaS :grimacing:

PR #505, opened 2023-06-28
Enable rubocop for development

mayorova:
I added RuboCop gem to development and test environments.
`.rubocop.yml` file was already there, but the gem was not in the Gemfile.

Also, the config was very outdated, so I updated the Cop names.

The 1st commit just makes the changes in the config (renaming obsolete ones, reordering according to the new namespaces etc.), but keeping them the same.

The 2nd commit cleans up the configs, removing all the cops that had `Enabled: false`, as it's redundant, given `AllCops:DisabledByDefault: true` setting. The rubocop output is the same for before and after the 2nd commit.

I also updated the `.codeclimate.yml` file, it was outdated too, so it was using some old Ruby. It's up-to-date now.


josemigallas (approved):

app/models/entry.rb
```diff
@@ -28,7 +28,7 @@ def process_entry
   end
 
   def self.for_model(model)
-    new(model: model, tenant: model.try!(:tenant))
+    new(model: model, tenant: model&.tenant)
```
akostadinov, line 31: @mayorova , this is not the same. `try!` should ignore any call errors while `&.tenant` will raise if `model` is not `nil` and `#tenant` method does not exist.

mayorova: @akostadinov I've checked this, and what you're saying is true for `.try`, on the other hand `.try!` does throw a `NoMethodError`, so its behavior is identical to `&.`

PR #506, opened 2023-06-29
Fix tenants controller test params wrapping

mayorova:
This was spotted by @jlledom here: https://github.com/3scale/zync/pull/502#discussion_r1244769959

The test is not correct, as by default (without `as: :json`) the request is sent with `application/x-www-form-urlencoded` Content-Type, and parameters wrapping is only enabled for JSON.

Also, I think we only support JSON (we probably need to reject requests with other content-types). So, the test should be for JSON also.

### DIFF
--- a/test/controllers/tenants_controller_test.rb
+++ b/test/controllers/tenants_controller_test.rb
@@ -4,9 +4,7 @@ require 'test_helper'
 class TenantsControllerTest < ActionDispatch::IntegrationTest
   test 'creating tenant' do
     assert_difference Tenant.method(:count) do
-      put tenant_url, params: {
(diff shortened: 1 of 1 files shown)


jlledom:
Are you sure we only support JSON? The wrapping won't work for other formats.

mayorova:
> Are you sure we only support JSON? The wrapping won't work for other formats.

@jlledom Well, I am pretty confident we do :) Because currently other formats does not work because of the missing wrapping (not related to Rails upgrade), and also we only really support Zync as part of the whole 3scale solution, and porta sends request to Zync only as JSON.

PR #507, opened 2023-06-29
Upgrade Que to version 1.4 (prep for v2)

mayorova:
Based on https://github.com/3scale/zync/pull/502

Gradual upgrade, based on the instructions here: https://github.com/que-rb/que/blob/master/CHANGELOG.md#200beta1-2022-03-24

>Recommended upgrade process:

>When using Que 2.x, a job enqueued with Ruby 2.7 will run as expected on Ruby 3. We recommend:

>Upgrade your project to the latest 1.x version of Que (1.3.1+)
IMPORTANT: adds support for zero downtime upgrade to Que 2.x, see changelog below
>Upgrade your project to Ruby 2.7 and Rails 6.x if it is not already
>Upgrade your project to Que 2.x but stay on Ruby 2.7
>IMPORTANT: You will need to continue to run Que 1.x workers until all jobs enqueued using Que 1.x (i.e. with a job_schema_version of 1) have been finished. See below
>Upgrade your project to Ruby 3

This is the first part, non-breaking, just adds a new column `job_schema_version` to the `que_jobs` table.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,7 @@ jobs:
 
       - run:
           name: Set up the DB
-          command: bundle exec bin/rails db:wait db:reset
+          command: bundle exec bin/rails db:wait db:setup
 
       - run:
           name: rails test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,8 +63,8 @@ jobs:
             - vendor/bundle
 
       - run:
-          name: rails db:setup
-          command: bundle exec bin/rails db:wait db:setup
+          name: Set up the DB
+          command: bundle exec bin/rails db:wait db:reset
 
       - run:
           name: rails test
--- a/Gemfile
+++ b/Gemfile
@@ -31,9 +31,7 @@ gem '3scale-api'
 
 gem 'bootsnap', '>= 1.4.4'
 
-# NOTE: do not upgrade to que v1.2 without upgrading Rails to v7, the deprecation introduced in 1.2.0 is resolved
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -63,8 +63,8 @@ jobs:
             - vendor/bundle
 
       - run:
-          name: rails db:setup
-          command: bundle exec bin/rails db:wait db:setup
+          name: Set up the DB
+          command: bundle exec bin/rails db:wait db:reset
```
akostadinov, line 67: why changed this? reset would try to remove existing database first.

akostadinov: The rest looks nice!

mayorova: I thought that that would be an easy way to resolve the "migrations are pending" error :grimacing: But it's not.
Adding `db:migrate` afterwards didn't help either :thinking: 

Ideas?

akostadinov: if the schema file is updated, then `db:setup` should apply it an no migrations should be missing
Maybe schema file version is wrong for some reason?

mayorova: Well, zync uses `structure.sql` instead of schema, and to be honest, I have no idea how it works and how it is related to migrations. 

And actually, I think there's something weird done with the filename (adding `-12` for any PG version > 11), I think I'll be fixing it in another PR.

akostadinov: It should be the content of `schema_migrations` table. Maybe we should drop support for postgres < v12 at some point.

mayorova: Yes, definitely, but well, first we need to migrate SaaS to at least 13, and upgrade 3scale-operator to use 13 as well (it uses 10 now)

akostadinov (approved):

PR #508, opened 2023-06-30
Add PostgreSQL 12 to tests and prevent from using unsupported versions

mayorova:
Add PostgreSQL 12 to tests, as we're still using it.

This PR also changes the names of the `structure.sql` file.

There have been two versions: `structure.sql` that was used for PostgreSQL version 10 and below, and `structure-12.sql` used for all other versions. This is very confusing as:
1. `structure-12.sql` would be used for PostgreSQL v13
2. `structure-12.sql` would also be used for PostgreSQL v11, BUT it would not work because of one statement unsupported in v11. There was in fact some discussion around it in Slack, but as we lost history, I don't think I'll be able to recover the details.

The differences between the two files are:

1. The following statement:
```
SET default_table_access_method = heap;
```

This appeared in PostgreSQL v12, and is not supported in earlier versions (this is why it would fail with PostgreSQL v11)

2. The following change in CREATE TRIGGER syntax:

```
CREATE TRIGGER xxx FOR EACH ROW EXECUTE FUNCTION yyy
```
is the preferred way to write the triggers, but in v10 only the following is supported:
```
CREATE TRIGGER xxx FOR EACH ROW EXECUTE PROCEDURE yyy
```

see https://www.postgresql.org/docs/release/11.0/
```
Also, writing FUNCTION is now preferred over writing PROCEDURE in CREATE OPERATOR and CREATE TRIGGER, because the referenced object must be a function not a procedure. However, the old syntax is still accepted for compatibility.
```

With this update, we:
- make it explicit that older versions of Postgres, as well as version 11 are not supported
- make it possible to get rid of `structure-10.sql` easily once we fully stop supporting it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,5 +90,5 @@ workflows:
       - build:
           matrix:
             parameters:
-              postgresql_image: [ "circleci/postgres:10-alpine-ram", "cimg/postgres:13.11" ]
+              postgresql_image: [ "circleci/postgres:10-alpine-ram", "cimg/postgres:12.15", "cimg/postgres:13.11" ]
       - docker-build
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM registry.access.redhat.com/ubi8/ruby-27
 
 USER root
 RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
-  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs --skip-broken -y shared-mime-info postgresql13 postgresql13-devel postgresql13-libs \
+  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs --skip-broken -y shared-mime-info postgresql13 postgresql13-libs \
   && dnf clean all \
   && rm -rf /var/cache/yum
 
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -28,7 +28,8 @@ ActiveRecord::Tasks::PostgreSQLDatabaseTasks.prepend(Module.new do
     clear_active_connections!
     establish_master_connection
     server_version = connection.select_value("SHOW server_version").to_i
-    suffix = server_version < 11 ? '' : '-12'
+    raise "PostgreSQL #{server_version} is not supported" if server_version < 10 || server_version == 11
+    suffix = server_version == 10 ? '-10' : ''
     ENV['SCHEMA'] ||= "db/structure#{suffix}.sql"
     clear_active_connections!
   end
--- a/db/structure-10.sql
+++ b/db/structure-10.sql
@@ -37,8 +37,6 @@ $$;
 
 SET default_tablespace = '';
 
-SET default_table_access_method = heap;
-
 --
 -- Name: que_jobs; Type: TABLE; Schema: public; Owner: -
 --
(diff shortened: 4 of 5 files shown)


.circleci/config.yml
```diff
@@ -7,20 +7,17 @@ jobs:
       - image: cimg/base:current
         environment:
           POSTGRES_CONTAINER_NAME: db
-          DATABASE_URL: postgresql://postgres:postgres@db:5432/zync
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgres
-          POSTGRES_DB: zync
+          DATABASE_URL: postgresql://postgres@localhost:5432/circle_test
           RAILS_ENV: production
+      - image: cimg/postgres:13.11
```
akostadinov, line 12:
```suggestion
      - image: cimg/postgres:13.11
        name: postgres
```

.circleci/config.yml
```diff
@@ -7,20 +7,17 @@ jobs:
       - image: cimg/base:current
         environment:
           POSTGRES_CONTAINER_NAME: db
-          DATABASE_URL: postgresql://postgres:postgres@db:5432/zync
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgres
-          POSTGRES_DB: zync
+          DATABASE_URL: postgresql://postgres@localhost:5432/circle_test
```
akostadinov, line 10:
```suggestion
          DATABASE_URL: postgresql://postgres@postgres:5432/circle_test
```

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM registry.access.redhat.com/ubi8/ruby-27
 
 USER root
 RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
-  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs --skip-broken -y shared-mime-info postgresql13 postgresql13-devel postgresql13-libs \
+  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs --skip-broken -y shared-mime-info postgresql13 postgresql13-libs \
```
jlledom, line 5: don't we need the `devel` package anymore?

jlledom (approved):

Dockerfile
mayorova: It seems that it is not needed :shrug: 
We do need the others, because we need postgres clients, which is provided by `postgresql13`, and `postgresql13-libs` is its dependency.

jlledom: `-devel` packages are required sometimes when you run `bundle install` locally, for instance, `mysql2` gem requires having `mysql-devel`. I tried to build the image locally and it succeeded, so it seems Postgres gem doesn't need the headers. It's fine for me then.

lib/tasks/db.rake
```diff
@@ -28,7 +28,8 @@ ActiveRecord::Tasks::PostgreSQLDatabaseTasks.prepend(Module.new do
     clear_active_connections!
     establish_master_connection
     server_version = connection.select_value("SHOW server_version").to_i
-    suffix = server_version < 11 ? '' : '-12'
+    raise "PostgreSQL #{server_version} is not supported" if server_version < 10 || server_version == 11
+    suffix = server_version == 10 ? '-10' : ''
```
akostadinov, line 32: why mark 11 as unsupported? I would suggest to leave it supported until we drop support for things below 12 altogether. But I'm fine if you feel confident nobody will suffer.

akostadinov (approved):

lib/tasks/db.rake
mayorova: The explanation is in the PR description - we never really officially supported v11, in fact, due to config differences, it will not work with the existing .sql files (we'd need to create a `structure-11.sql`)

PR #510, opened 2023-07-06
Ruby upgrade to v3

mayorova:
Some issues encountered in the upgrade:

1. The dependency `k8s-ruby`, or rather its dependency `recursive-open-struct` uses deprecated arguments passing here: https://github.com/aetherknight/recursive-open-struct/blob/main/lib/recursive_open_struct.rb#L188

It seems that it kind of supports Ruby 3 https://github.com/aetherknight/recursive-open-struct/blob/main/.travis.yml#L15

As well as `k8s-ruby` claims that it supports Ruby 3.1 https://github.com/k8s-ruby/k8s-ruby/commit/30d4db1b8ad991660098c89e55b35ee18aa216c9

But probably it's an edge case...

With Ruby 2.7 there is a deprecation warning: 
```

/home/dmayorov/.asdf/installs/ruby/2.7.6/lib/ruby/gems/2.7.0/gems/recursive-open-struct-1.1.3/lib/recursive_open_struct.rb:188: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

And with Ruby 3 it fails with ArgumentError.

For now the solution has been changing the arguments of `Integration::KubernetesService::Route.initialize`. But we can review if that's a reasonable fix.

Maybe we can also look into using another library instead of `k8s-ruby`, for example, [Kubernetes documentation](https://kubernetes.io/docs/reference/using-api/client-libraries/#officially-supported-kubernetes-client-libraries) suggests https://kubernetes.io/docs/reference/using-api/client-libraries/ as officially supported (also, not sure if it's well-maintained).

2. Test suite was getting either stuck (in CircleCI), or failing with the following error:
```
ThreadError: deadlock; lock already owned by another fiber belonging to the same thread
```
After looking into it, it turned out to be caused by two tests that explicitly used Fibers.

Apparently, this does not work well within a transaction. There is a fix applied here: https://github.com/rails/rails/pull/46553 (that we confirmed would fix the tests for us).

However, it's not in the latest released version of Rails (not in 7.0.6 at least). So, we're skipping these tests for now, until we upgrade Rails to a version where this is fixed.

3. There were some issues with the `ubi9/ruby-31` base image, for example, `rdoc` is not present as a default Gem. @akostadinov railsed an issue in Bugzilla for this: https://bugzilla.redhat.com/show_bug.cgi?id=2221938

Also, both `irb` and `rdoc` were not available by default and had to be installed explicitly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,9 +42,7 @@ jobs:
       # Restore bundle cache
       - restore_cache:
           keys:
-            - zync-bundle-v2-{{ arch }}-{{ checksum "Gemfile.lock" }}
-            - zync-bundle-v2-{{ arch }}-{{ .Branch }}
-            - zync-branch-v2-{{ arch }}-master
+            - zync-bundle-v2-{{ .Environment.CACHE_VERSION }}-{{ arch }}-{{ checksum "Gemfile.lock" }}
 
       - run:
           name: bundle install
@@ -59,7 +57,7 @@ jobs:
           command: BUNDLE_WITHOUT=development:test bundle exec bin/rails runner --environment=production 'puts Rails.env'
 
       - save_cache:
-          key: zync-bundle-v2-{{ arch }}-{{ checksum "Gemfile.lock" }}
+          key: zync-bundle-v2-{{ .Environment.CACHE_VERSION }}-{{ arch }}-{{ checksum "Gemfile.lock" }}
           paths:
             - vendor/bundle
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,8 @@ jobs:
         type: string
     working_directory: /opt/app-root/zync
     docker:
-      - image: registry.access.redhat.com/ubi7/ruby-27
+      # A Zync image tagged manually for CI tests
+      - image: quay.io/3scale/zync:ci-builder
       - image: << parameters.postgresql_image >>
     environment:
         RAILS_ENV: test
@@ -48,7 +49,7 @@ jobs:
       - run:
           name: bundle install
           command: |
-            gem install bundler --version=$(grep -A 1 "BUNDLED WITH" Gemfile.lock | tr -d ' '| tail -n 1)
+            gem install bundler --version=$(grep -A 1 "BUNDLED WITH" Gemfile.lock | tr -d ' '| tail -n 1) --no-document
             bundle config --local force_ruby_platform true
             bundle config set --local deployment 'true'
             bundle config set --local path 'vendor/bundle'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,7 +68,7 @@ jobs:
       - run:
           name: rails test
           command: |
-            circleci tests glob "test/**/*_test.rb" | circleci tests run --command="xargs bundle exec rails test" --verbose --split-by=timings
+            circleci tests glob "test/**/*_test.rb" | circleci tests run --command="xargs bundle exec rake test TESTOPTS='-v'" --verbose --split-by=timings
       - run:
           name: license_finder
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,7 +68,7 @@ jobs:
       - run:
           name: rails test
           command: |
-            bundle exec bin/rails test $(circleci tests glob "test/**/*_test.rb" | circleci tests split --split-by=timings)
+            circleci tests glob "test/**/*_test.rb" | circleci tests run --command="xargs bundle exec rails test" --verbose --split-by=timings
       - run:
           name: license_finder
           command: |
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,13 @@
 FROM registry.access.redhat.com/ubi9/ruby-31
 
 USER root
-RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
-    && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs --skip-broken -y shared-mime-info postgresql13 postgresql13-libs \
+RUN dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y shared-mime-info postgresql rubygem-irb rubygem-rdoc \
     && dnf clean all \
     && rm -rf /var/cache/yum
+
+# worksround https://bugzilla.redhat.com/show_bug.cgi?id=2221938
+RUN ln -s /usr/share/gems/gems/rdoc-6.4.0/lib/rdoc.rb /usr/share/ruby/ \
+    ln -s /usr/share/gems/gems/rdoc-6.4.0/lib/rdoc /usr/share/ruby/
 
 USER default
 WORKDIR ${APP_ROOT}
@@ -14,8 +17,7 @@ COPY --chown=default:root Gemfile* ./
 RUN BUNDLER_VERSION=$(awk '/BUNDLED WITH/ { getline; print $1 }' Gemfile.lock) \
     && gem install bundler --version=$BUNDLER_VERSION --no-document
 
-RUN bundle config build.pg --with-pg-config=/usr/pgsql-13/bin/pg_config \
-    && bundle config set --local deployment 'true' \
+RUN bundle config set --local deployment 'true' \
     && bundle config set --local path 'vendor/bundle' \
     && bundle install --jobs $(grep -c processor /proc/cpuinfo) --retry 3
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y s
     && dnf clean all \
     && rm -rf /var/cache/yum
 
-# worksround https://bugzilla.redhat.com/show_bug.cgi?id=2221938
+# Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2221938
 RUN ln -s /usr/share/gems/gems/rdoc-6.4.0/lib/rdoc.rb /usr/share/ruby/ \
     ln -s /usr/share/gems/gems/rdoc-6.4.0/lib/rdoc /usr/share/ruby/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM registry.access.redhat.com/ubi8/ruby-27
+FROM registry.access.redhat.com/ubi9/ruby-31
 
 USER root
 RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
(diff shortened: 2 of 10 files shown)


Gemfile
```diff
@@ -56,6 +56,8 @@ gem 'yabeda-rails'
 gem 'yabeda-prometheus', '~> 0.6.1'
 gem 'yabeda-puma-plugin'
 
+gem 'webrick'
+
```
akostadinov, line 60: we need this?

mayorova: Yes, this gem was excluded from Ruby since 3.0

https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

> The following libraries are no longer bundled gems or standard libraries. Install the corresponding gems to use these features.
>  - sdbm
>  - webrick
>  - net-telnet
>  - xmlrpc

.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         type: string
     working_directory: /opt/app-root/zync
     docker:
-      - image: registry.access.redhat.com/ubi7/ruby-27
+      - image: quay.io/3scale/zync:test-base
```
mayorova, line 32: The problem with the `registry.access.redhat.com/ubi9/ruby-31` image is that it lacks PostgreSQL libraries, and I didn't manage to install them with `dnf` as a CircleCI run step.

So, **temporarily** (just to check what other issues I may encounter) I replaced the ubi image with a custom one, built from the following Dockerfile:
```
FROM registry.access.redhat.com/ubi9/ruby-31

USER root
    && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs --skip-broken -y shared-mime-info postgresql13 postgresql13-libs \
    && dnf clean all \
    && rm -rf /var/cache/yum

USER default
```

.circleci/config.yml
```diff
@@ -41,14 +41,14 @@ jobs:
       # Restore bundle cache
       - restore_cache:
           keys:
-            - zync-bundle-v2-{{ arch }}-{{ checksum "Gemfile.lock" }}
-            - zync-bundle-v2-{{ arch }}-{{ .Branch }}
-            - zync-branch-v2-{{ arch }}-master
+            - zync-bundle-v2-{{ .Environment.CACHE_VERSION }}-{{ arch }}-{{ checksum "Gemfile.lock" }}
```
mayorova, line 44: This was also just to try things out. It's similar to what we do in portal. If it bothers here, I can remove.

Gemfile
```diff
@@ -68,6 +70,9 @@ group :development, :test do
   gem 'rubocop', require: false
   gem 'rubocop-performance', require: false
   gem 'rubocop-rails', require: false
+
+  gem 'irb'
```
mayorova, line 74: Apparently, `ubi9/ruby-31` image does not have these installed.

Also, I was testing the image built by the `Dockerfile` locally, and even when I install `rubygem-irb` and `rubygem-rdoc` dnf packages in the Dockerfile, rails can't start because it doesn't find them. Weird! Probably, I'm just doing something wrong.

INSTALL.md
```diff
@@ -7,13 +7,13 @@
...
 
-## Install dependencies.
+## Install dependencies
 
 To run Zync you need access to a running [PostgreSQL](https://www.postgresql.org) server. You can install one with your operating system package
 manager, as a container or run it remotely.
 
 The minimum requirement for the machine running Zync is to have
- - Ruby 2.7.x
+ - Ruby 3.x
```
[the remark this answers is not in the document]
mayorova: Yeah, will update once we for sure use this one.

app/services/integration/kubernetes_service.rb
```diff
@@ -124,7 +124,7 @@ def labels_for_provider(entry)
   end
 
   class Route < K8s::Resource
-    def initialize(attributes, **options)
+    def initialize(attributes, options = {})
```
mayorova, line 127: I don't like this hack... Some info on why it is here is in the PR description.

Maybe we need to switch to another k8s client library :shrug:

Gemfile
akostadinov: It is only the `rdoc` that creates the troubles. Looking into it.

Gemfile
akostadinov: I wonder if we can use puma for dev and prod as well. I assume we run puma in production. Or not?

mayorova: It is a dependency for `yabeda-prometheus`, I left a comment.
It's possible to use a different web server, but we can keep it as it was for now.

test/services/incoming_notification_service_test.rb
```diff
@@ -35,6 +35,10 @@ def setup
   class LockingTest < ActiveSupport::TestCase
     self.use_transactional_tests = false
 
+    def setup
+      skip "see https://github.com/rails/rails/pull/46553" if Rails.version <= Gem::Version.new('7.0.6')
```
mayorova, line 39: it is not in the beginning of the test itself, because of this block:
```
    ensure
      assert_nil fiber.resume
    end
```

.circleci/config.yml
```diff
@@ -29,7 +29,8 @@ jobs:
         type: string
     working_directory: /opt/app-root/zync
     docker:
-      - image: registry.access.redhat.com/ubi7/ruby-27
+      # A Zync image tagged manually for CI tests
+      - image: quay.io/3scale/zync:ci-builder
```
jlledom, line 33: Is this base image still temporary?

Dockerfile
```diff
@@ -1,10 +1,13 @@
...
 USER root
-RUN rpm -Uvh https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
-  && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs --skip-broken -y shared-mime-info postgresql13 postgresql13-libs \
-  && dnf clean all \
-  && rm -rf /var/cache/yum
+RUN dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y shared-mime-info postgresql rubygem-irb rubygem-rdoc \
+    && dnf clean all \
+    && rm -rf /var/cache/yum
+
+# worksround https://bugzilla.redhat.com/show_bug.cgi?id=2221938
```
jlledom, line 8: Typo: `worksround`

Dockerfile
```diff
@@ -1,10 +1,13 @@
...
+RUN ln -s /usr/share/gems/gems/rdoc-6.4.0/lib/rdoc.rb /usr/share/ruby/ \
+    ln -s /usr/share/gems/gems/rdoc-6.4.0/lib/rdoc /usr/share/ruby/
```
jlledom, line 10: Wouldn't be better to add `rdoc` to the Gemfile? I know it's not nice but it's nicer than this, IMO.

app/adapters/three_scale/api/instrumented_http_client.rb
```diff
@@ -5,8 +5,8 @@
 # Custom HTTP Client for 3scale API client with added instrumentation.,
 class ThreeScale::API::InstrumentedHttpClient < ThreeScale::API::HttpClient
 
-  def initialize(**)
-    super
+  def initialize(...)
+    super(...)
```
jlledom, line 9: Why this?

test/services/incoming_notification_service_test.rb
jlledom: What if you use threads instead of fibers? wouldn't the test still be valid?

Dockerfile
akostadinov: I think it is worse to put bloat into the project instead of fixing the image. We don't depend on rdoc and it seems wrong to me to include it because of broken ruby distribution.

.circleci/config.yml
mayorova: No, me and @akostadinov discussed this and decided that this is a reasonable approach.

Basically, the problem is that `ubi9/ruby-31` misses some dependencies required to run zync: irb, rdoc on Ruby side, but also PostgreSQL client.

So, we considered various solutions:

1. Create a separate repository for building the image for Zync CI, like porta has: https://github.com/3scale/system-builder
We considered this is an overkill

2. (this approach) Just use a manually tagged Zync image for CI

3. Create a separate Dockerfile in zync repo for building an image for CI, and push it to quay. Potentially, CI itself could build this image

4. Make the docker build step (from the existing Dockerfile) a dependency for the rest of the tests.

2 seemed the most simple and effective.

test/services/incoming_notification_service_test.rb
akostadinov: Probably. But it will be more messy to synchronize. If you remember 3scale/porta#3403
So we think it is reasonable to wait until next rails bump to re-enable them

app/services/integration/kubernetes_service.rb
akostadinov: I think the point was that gem is outdated and a potential maintenance issue going forward.

test/services/incoming_notification_service_test.rb
mayorova: Yes, just trying to avoid extra work to rewrite these tests completely.

jlledom (dismissed):

test/services/incoming_notification_service_test.rb
```diff
@@ -36,7 +36,7 @@ class LockingTest < ActiveSupport::TestCase
     self.use_transactional_tests = false
 
     def setup
-      skip "see https://github.com/rails/rails/pull/46553" if Rails.version <= Gem::Version.new('7.0.6')
+      skip 'see https://github.com/rails/rails/pull/46553' if Rails.version <= Gem::Version.new('7.0.6')
```
akostadinov, line 39: why use single quotes? Not a problem, just asking. I thought double quotes are generally preferred always unless they enclose other double quotes.

mayorova: I think both are equally correct, as long as they are consistent: https://github.com/rubocop/ruby-style-guide#consistent-string-literals

I think this project mostly uses "default single quote" style, where double quote is only used when interpolation is needed. I personally also prefer this one.

.circleci/config.yml
```diff
@@ -69,7 +70,9 @@ jobs:
       - run:
           name: rails test
           command: |
-            bundle exec bin/rails test $(circleci tests glob "test/**/*_test.rb" | circleci tests split --split-by=timings)
+            TESTS=$(circleci tests glob "test/**/*_test.rb" | circleci tests split --split-by=timings)
+            echo $TESTS
+            bundle exec bin/rails test $TESTS
```
akostadinov, line 75:
```suggestion
            circleci tests glob "test/**/*_test.rb" | circleci tests run --command="xargs bundle exec rails test" --verbose --split-by=timings
```

mayorova: Yes, I'll do this also. It's just that I am trying to pinpoint the issue with tests timing out :confused: 
I didn't see it before, so I guess this is the result of Ruby upgrade. I hope it's just the tests, but would like to understand why it is happening, so trying to figure out whether it's a specific test that times out.

jlledom (approved):

PR #514, opened 2023-07-31
:broom: Remove bugsnag-capistrano dependency and update licenses

mayorova:
We are not using capistrano for deployments, so just cleaning up unneeded deps.

Also, when updating the licenses file I realized that it was not updated after Rails upgrade.

Probably we need to generate it in the container build, like we do for Porta...


akostadinov (approved):
+1 for generating licenses in image like porta.

PR #515, opened 2023-08-01
Remove bugnsag-capistrano rake command

mayorova:
A follow-up for https://github.com/3scale/zync/pull/514


mayorova:
@akostadinov The title refers to the `bugnsag-capistrano` gem that provided the rake task `bugsnag:deploy` that I am deleting in this PR.
The gem itself was deleted in https://github.com/3scale/zync/pull/514, so this task wouldn't work anyway.

Yes, indeed this actually has nothing to do with capistrano, it was just that the rake task was happened to be provided by this capistrano-related gem. But, as far as I know, we don't use this rake task anyway, so... :wave:

PR #525, opened 2024-04-22
Upgrade Nokogiri to v1.16.4

mayorova:
Supersedes https://github.com/3scale/zync/pull/523 and https://github.com/3scale/zync/pull/519

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -180,7 +180,7 @@ GEM
       rack (>= 1.1.3)
     method_source (1.0.0)
     mini_mime (1.1.2)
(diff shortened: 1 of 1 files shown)


akostadinov (approved):

akostadinov:
btw we may look at using the system libxml version so that we don't have to update nokogiry so often. I think many times vulnerabilities are in the c library and not on the ruby side. Looking at last few versions, it's all about that.

See https://github.com/3scale/porta/blob/8aef7ac1710256cf34e8573a9e7441ca5b9df288/openshift/system/Dockerfile#L14

PR #559, opened 2025-07-30
Using plain ubi9 base image

mayorova:
`registry.access.redhat.com/ubi9/ruby-31` is deprecated...


akostadinov (dismissed):

akostadinov:
MErging this thing, not sure codeclimate works at all anymore

PR #567, opened 2025-10-08
test with all supported PostgreSQL versions

akostadinov:
Postgres 10 is just a nuisance already and long unsupported. So removing this file. And added all supported PostgreSQL versions in the test matrix so we know they are compatible.


lib/tasks/db.rake
akostadinov, line 1: Now we have only one schema for all versions, I don't see why we need all this extra logic.

PR #591, opened 2026-04-29
Upgrade postgres to version 15

mayorova:
Apparently, version 13 was installed in the upstream images, and it didn't work with postgres v15.

The initial attempt was just to enable the module with the newer version, as we do it in midstream:
```
dnf -y module enable postgresql:15
```

However, it turns out that this module is not available in UBI9 repos, only in RHEL9 :(

So, I tried upgrading to UBI10 instead. Apparently, in UBI10 modules are not available anymore:
```
WARNING: modularity is deprecated, and functionality will be removed in a future release of DNF5.
```

But I verified that actually the `ruby` version is `3.3` (the one we need), and `postgresql` is `16.13`. It's one version above the one we support, but it seems that PG client can work fine with older server versions (not the other way around) - although I'm struggling to find official statement from PostgreSQL on this.
I hope this will work though for the upstream image - and also, this way we can easier upgrade to supporting PostgreSQL 16 :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9:9.6
+FROM registry.access.redhat.com/ubi9:9.7
 
 ENV RUBY_MAJOR_VERSION=3 \
     RUBY_MINOR_VERSION=3 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -7,8 +7,10 @@ ENV RUBY_VERSION="${RUBY_MAJOR_VERSION}.${RUBY_MINOR_VERSION}"
 
 USER root
 
-RUN dnf -y module enable ruby:${RUBY_VERSION} \
-    && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y shared-mime-info make automake gcc gcc-c++ postgresql git ruby-devel rubygem-irb rubygem-rdoc glibc-devel libpq-devel libyaml-devel xz \
+RUN dnf update --setopt=install_weak_deps=0 --setopt=tsflags=nodocs -y \
+    && dnf -y module enable ruby:${RUBY_VERSION} \
+    && dnf -y module enable postgresql:15 \
+    && dnf install --setopt=skip_missing_names_on_install=False,tsflags=nodocs -y shared-mime-info postgresql rubygem-irb rubygem-rdoc make automake gcc gcc-c++ git ruby-devel glibc-devel libpq-devel libxml2-devel libxslt-devel libyaml-devel xz \
```
mayorova, line 8: This line is just copy-paste from midstream for easier tracking. Most of the libs are the same.

jlledom (dismissed):

Dockerfile
akostadinov: you added `libxslt-devel` and `libyaml-devel`, any particular reason?

mayorova: The reasons are explained above - I just copy-pasted this line from midstream.

probably something because of this: https://github.com/3scale/zync/pull/572#discussion_r2607051008

I really didn't want to figure out the details and remove repos just in case they are not needed.

akostadinov: You could just try to build it and see. Porta has different dependencies than zync. I think we generally don't install packages that are not needed and clean-up what's not needed.

akostadinov:
I'd suggest sticking to the ubi9 default as well in the production images. They should be compatible with postgres 15 server.

Dockerfile
mayorova: That comment I referenced is for zync.
I just wanted it to be closer to production image. If we don't need it - we can remove it in both, probably.

mayorova:
Closing in favor of https://github.com/3scale/zync/pull/592

